PR #1, opened 2021-11-15
Prototype of streaming departure message handling

DavidGregory084:
This is a prototype of how we could handle XML declaration data messages using Akka streams.

This is becoming increasingly important as we learn about the volumes on P4 and the expected increase in message sizes in NCTS P5.


drnaylor-oc:
Closing this as we've now built up our services and I think we have everything that was useful from here. Thanks for this original work David, it definitely helped us with some of the streaming concepts!

PR #3, opened 2022-05-23
CTCP-531: Create service to extract movement EORI

drnaylor-oc:
This is an initial pass, will need to extract the generated date. Also need to check to see if messageSender is indeed the EORI.

There is also a question about DeclarationData - I'm considering making that purely return extracted data and let the controller pass the non-extracted data to where it needs to go.

Looking to get something like this merged so that the signature is available to @liandra-giacomine and @ardwalker for their respective tickets.


liandra-giacomine (approved):

RobertLaverick (changes requested):
As discussed, I'll happily approve this once Partial Declaration Data disappears

RobertLaverick (approved):

PR #4, opened 2022-05-23
Added DeparturesController with post function

liandra-giacomine:


app/uk/gov/hmrc/transitmovements/controllers/DeparturesController.scala
```diff
@@ -0,0 +1,93 @@
...
+          BaseError.internalServiceError(cause = Some(thr))
+      }
+      .flatMap {
+        temporaryFile =>
+          val source = request.body.alsoTo(FileIO.toPath(temporaryFile.path))
+          onSucceed(temporaryFile, source).bimap(deleteFile(temporaryFile), deleteFile(temporaryFile))
+      }
+
+  def translateParseError(parseError: ParseError): BaseError = parseError match {
+    case NoElementFound(element)       => BaseError.badRequestError(s"Element $element not found in departure declaration")
```
drnaylor-oc, line 74: Although we're only doing departure declarations at this stage, these parse errors are going to be common - so make these messages more generic.

app/uk/gov/hmrc/transitmovements/controllers/errors/ErrorCode.scala
```diff
@@ -0,0 +1,58 @@
...
+    errorCode =>
+      errorCodes
+        .find(
+          value => value.code == errorCode.asInstanceOf[JsString].value
+        )
+        .map(
+          errorCode => JsSuccess(errorCode)
+        )
+        .getOrElse(JsError())
+  }
```
drnaylor-oc, line 57: Perhaps use pattern matching here instead, something like:

```scala
Reads {
    case JsString(errorCode) =>
      // .. code here
    case _ => JsError()
  }
```

drnaylor-oc (commented):
Some musings.

RobertLaverick (approved):

PR #5, opened 2022-05-23
CTCP-531: Add graph for combining multiple entries, assuming they're all single entries

drnaylor-oc:
This now returns the preparation date/time as the generated time.

Also creates a `ParseResult[A]` type alias for `Either[ParseResult, A]` to shorten the code a little.


RobertLaverick (changes requested):

RobertLaverick (approved):
I'm going to write a spike to look at isolating the flows and making some of this more generic to see if that is actually technical debt.

For now, lets get this in. :shipit:

RobertLaverick (approved):
thank you for fixing test

PR #9, opened 2022-05-26
Added ErrorTranslator trait to DeparturesController

liandra-giacomine:
Implicitly converting errors (e.g. ParseError) in EitherTs into BaseError

### DIFF
--- a/app/uk/gov/hmrc/transitmovements/controllers/errors/ErrorTranslator.scala
+++ b/app/uk/gov/hmrc/transitmovements/controllers/errors/ErrorTranslator.scala
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2022 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.transitmovements.controllers.errors
+
+import cats.data.EitherT
+import uk.gov.hmrc.transitmovements.services.errors.ParseError
+import uk.gov.hmrc.transitmovements.services.errors.ParseError.BadDateTime
+import uk.gov.hmrc.transitmovements.services.errors.ParseError.NoElementFound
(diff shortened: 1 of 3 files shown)


app/uk/gov/hmrc/transitmovements/controllers/errors/ErrorTranslator.scala
```diff
@@ -0,0 +1,46 @@
...
+import uk.gov.hmrc.transitmovements.services.errors.ParseError
+import uk.gov.hmrc.transitmovements.services.errors.ParseError.BadDateTime
+import uk.gov.hmrc.transitmovements.services.errors.ParseError.NoElementFound
+import uk.gov.hmrc.transitmovements.services.errors.ParseError.TooManyElementsFound
+import uk.gov.hmrc.transitmovements.services.errors.ParseError.Unknown
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Future
+
+trait ErrorTranslator {
```
drnaylor-oc, line 29: We should add unit tests for this trait specifically.

app/uk/gov/hmrc/transitmovements/controllers/errors/ErrorTranslator.scala
```diff
@@ -0,0 +1,43 @@
...
+
+  implicit def translateError[E, A](either: EitherT[Future, E, A]): EitherT[Future, BaseError, A] =
```
drnaylor-oc, line 31: Preface: there is nothing wrong with what you've done here, but I am concerned about the use of an `implict def` like this.

I do like this in that it cleans up the for comprehension... but we do end up with the situation where we've hidden some logic due to this implicit conversion which could trip up someone new starting on the project. To be fair, [I've used them in the testing repo](https://github.com/hmrc/ctc-traders-api-tests/blob/9dd6f36bbbe5dcfbd3e1058a2f0351a185f21e9b/src/test/scala/uk/gov/hmrc/ctc/tests/tags/TestTag.scala#L10), but that was to create a mini-DSL for their tagging and is to just make a type conversion implicit rather than explicit.

Their use is often discouraged too - in fact, one of the creators of Scala, Martin Odesky [says they're evil and should be removed from Scala](https://contributors.scala-lang.org/t/can-we-wean-scala-off-implicit-conversions/4388) - though that is in the context of "better alternatives now exist".

Important question that I think we all should - do we want to use implicit conversions in the code like this - especially as this does some processing?

RobertLaverick: I'd lean against using them in this circumstance, personally, but as you said, there is nothing wrong with it

drnaylor-oc: Yeah - most of this is still needed though so I think we're gonna go for a typeclass based approach, which has the added bonus of making a lack of conversion a compile time error instead of a runtime error.

PR #21, opened 2022-08-04
CTCP-774 and CTCP-842

RobertLaverick:
Added support for getting departures (without messages) and getting individual messages


app/uk/gov/hmrc/transitmovements/models/DepartureWithoutMessages.scala
```diff
@@ -0,0 +1,54 @@
...
+import java.time.OffsetDateTime
+
+case class DepartureWithoutMessages(
+  _id: DepartureId,
+  enrollmentEORINumber: EORINumber,
+  movementEORINumber: EORINumber,
+  movementReferenceNumber: Option[MovementReferenceNumber], // optional pending MRN allocation
+  created: OffsetDateTime,
+  updated: OffsetDateTime
+) {}
```
drnaylor-oc, line 31: nitpick: we don't need braces here.

app/uk/gov/hmrc/transitmovements/services/DeparturesFactory.scala
```diff
@@ -44,68 +46,63 @@ import scala.concurrent.ExecutionContext.Implicits.global
...
+          created = OffsetDateTime.ofInstant(clock.instant, ZoneOffset.UTC),
+          updated = OffsetDateTime.ofInstant(clock.instant, ZoneOffset.UTC),
+          messages = NonEmptyList.one(createDeclarationMessage(declarationData, message))
+        )
+    }
 
-  private def createDeclarationMessage(declarationData: DeclarationData, messageBody: Option[String]): MovementMessage =
-    MovementMessage(
-      id = MovementMessageId(ShortUUID.next(clock, random)),
+  //To Consider: Refactor to a MessageFactory?
```
drnaylor-oc, line 84: We're expecting the message object to be the same across departures and arrivals, so yes, I think that would be wise - but this would need `generationDate` and `messageType` to become parameters.

Is it reasonable to do it now while fresh in heads, or is this a note we need to carry forward?

drnaylor-oc (commented):
I don't have much to say about this - but I did help with some of it soooo... I'll just offer a thought and a nitpick.

app/uk/gov/hmrc/transitmovements/services/DeparturesFactory.scala
RobertLaverick: I think it's best to take it forward to when we persist other message types. It'll need to be usable in both this scenario and when we create other messages, and it'll be clearer at that time.

drnaylor-oc (approved):

PR #22, opened 2022-08-15
CTCP-744

RobertLaverick:
Added getDepartureMessageIds Endpoint

### DIFF
--- a/app/uk/gov/hmrc/transitmovements/repositories/DeparturesRepository.scala
+++ b/app/uk/gov/hmrc/transitmovements/repositories/DeparturesRepository.scala
@@ -29,7 +29,6 @@ import org.mongodb.scala.model.IndexModel
 import org.mongodb.scala.model.IndexOptions
 import org.mongodb.scala.model.Indexes
 import play.api.Logging
-import play.api.libs.json.Json
 import uk.gov.hmrc.mongo.MongoComponent
 import uk.gov.hmrc.mongo.play.json.Codecs
 import uk.gov.hmrc.mongo.play.json.PlayMongoRepository
@@ -54,13 +53,6 @@ import scala.util.Failure
 import scala.util.Success
 import scala.util.Try
 import scala.util.control.NonFatal
-
-case class GetDepartureMessageIdsDTO(result: NonEmptyList[MessageId])
-
-object GetDepartureMessageIdsDTO extends CommonFormats with ModelFormats {
-
-  implicit val format = Json.format[GetDepartureMessageIdsDTO]
-}
 
 @ImplementedBy(classOf[DeparturesRepositoryImpl])
 trait DeparturesRepository {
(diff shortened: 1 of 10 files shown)


app/uk/gov/hmrc/transitmovements/repositories/DeparturesRepository.scala
```diff
@@ -112,6 +128,30 @@ class DeparturesRepositoryImpl @Inject() (
...
   }
 
+  def getDepartureMessageIds(eoriNumber: EORINumber, departureId: DepartureId): EitherT[Future, MongoError, Option[NonEmptyList[MessageId]]] = {
+
+    val selector = mAnd(mEq("_id", departureId.value), mEq("enrollmentEORINumber", eoriNumber.value))
+
+    val aggregates =
+      Seq(
+        Aggregates.filter(selector),
+        Aggregates.group(null, Accumulators.push("result", "$messages.id")),
```
derekspalmer, line 138: I know it works but think I need an explanation on this .group  to fully understand, please!

app/uk/gov/hmrc/transitmovements/repositories/DeparturesRepository.scala
```diff
@@ -50,11 +55,19 @@ import scala.util.Success
 import scala.util.Try
 import scala.util.control.NonFatal
 
+case class GetDepartureMessageIdsDTO(result: NonEmptyList[MessageId])
```
derekspalmer, line 58: Are these in the best place?

derekspalmer (approved):

app/uk/gov/hmrc/transitmovements/repositories/DeparturesRepository.scala
drnaylor-oc: I can see why it's in this file (it is private to it), but perhaps it should be a nested class to enforce the hierarchy, rather than here.

app/uk/gov/hmrc/transitmovements/repositories/DeparturesRepository.scala
drnaylor-oc: https://www.mongodb.com/docs/manual/reference/operator/aggregation/group/#std-label-null-example

It's using the aggregation to create a new array that has all the message IDs pushed to it. It basically takes each object, gets the ID field in the messages subdocument, pushes them to a new array named "result", and returns only that. Though there should only be one document at this stage and so there may be a projection may work, there were issues in getting the `replaceRoot` to work with an array, so this was easier.

The below two methods turns the document with a results array into an arrray of documents with one ID each, I believe.

It might be worth opening MongoDB Compass or your favourite Mongo workbench and seeing how it fits together.

app/uk/gov/hmrc/transitmovements/repositories/DeparturesRepository.scala
drnaylor-oc: Ah, okay. I see why it isn't an inner class.

Note that this is added as a codec to the repository in its constructor. Because of this, this _cannot_ be an inner class because you end up with a chicken and egg scenario - and Scala doesn't like it.

Thinking ahead, this class will likely be used in he arrivals repository too. So, I'll move this into its own file.

drnaylor-oc:
I just ran this through the code from my PR and it worked successfully. However, as I have now touched this PR, I need someone else to review.

derekspalmer (approved):

PR #23, opened 2022-08-17
Ctcp 962

derekspalmer:


ardwalker (dismissed):

it/uk/gov/hmrc/transitmovements/repositories/DeparturesRepositorySpec.scala
```diff
@@ -142,4 +144,60 @@ class DeparturesRepositorySpec
     val result = await(repository.getDepartureMessageIds(EORINumber("ABC"), DepartureId("XYZ")).value)
     result.right.get.isEmpty should be(true)
   }
+
+  "getDepartureIds" should
+    "return a list of departure ids for the supplied EORI sorted by last updated, latest first" in new DBSetup {
```
drnaylor-oc, line 149: I get why you're doing this `new DBSetup` thing, in principle I have no issue with the pattern. However, you're dropping everything and recreating it each time, which is just a waste of time. I'd prefer to see the database setup once and torn down once.

derekspalmer: Agree it takes longer, but It was to maintain clean tests. 
Will change to as suggested.

derekspalmer: Change to instantiate once dbset up once.

drnaylor-oc (approved):

PR #24, opened 2022-08-18
CTCP-829: Add ability to filter message IDs by received date

drnaylor-oc:
Also added an OffsetDateTime reads/writes to enable support for Mongo Date objects via the LocalDateTime formats.

### DIFF
--- a/app/uk/gov/hmrc/transitmovements/repositories/DeparturesRepository.scala
+++ b/app/uk/gov/hmrc/transitmovements/repositories/DeparturesRepository.scala
@@ -23,6 +23,7 @@ import com.google.inject.ImplementedBy
 import com.mongodb.client.model.Filters.{ne => mNe}
 import com.mongodb.client.model.Filters.{and => mAnd}
 import com.mongodb.client.model.Filters.{eq => mEq}
+import com.mongodb.client.model.Filters.{gte => mGte}
 import org.mongodb.scala.bson.BsonDocument
 import org.mongodb.scala.bson.conversions.Bson
 import org.mongodb.scala.model._
@@ -39,9 +40,13 @@ import uk.gov.hmrc.transitmovements.models.Message
 import uk.gov.hmrc.transitmovements.models.MessageId
 import uk.gov.hmrc.transitmovements.models.formats.CommonFormats
 import uk.gov.hmrc.transitmovements.models.formats.ModelFormats
+import uk.gov.hmrc.transitmovements.repositories.DeparturesRepositoryImpl.EPOCH_TIME
 import uk.gov.hmrc.transitmovements.services.errors.MongoError
 import uk.gov.hmrc.transitmovements.services.errors.MongoError._
 
+import java.time.LocalDateTime
+import java.time.OffsetDateTime
+import java.time.ZoneOffset
 import java.util.concurrent.TimeUnit
 import javax.inject.Inject
 import scala.concurrent._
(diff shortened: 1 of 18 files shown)


derekspalmer (dismissed):

app/uk/gov/hmrc/transitmovements/repositories/DeparturesRepository.scala
```diff
@@ -120,15 +134,26 @@ class DeparturesRepositoryImpl @Inject() (
...
+    receivedSince: Option[OffsetDateTime]
+  ): EitherT[Future, MongoError, Option[NonEmptyList[MessageId]]] = {
+
+    val selector = mAnd(
+      mEq("_id", departureId.value),
+      mEq("enrollmentEORINumber", eoriNumber.value)
+    )
 
-    val selector = mAnd(mEq("_id", departureId.value), mEq("enrollmentEORINumber", eoriNumber.value))
+    val dateTimeSelector = mGte("messages.received", receivedSince.map(_.toLocalDateTime).getOrElse(EPOCH_TIME))
```
ardwalker, line 148: Would EPOCH_TIME potentially result in many messages being returned? Are messages actually purged from the database after 31 days? If so, I guess this will never be a problem.

it/uk/gov/hmrc/transitmovements/repositories/DeparturesRepositorySpec.scala
```diff
@@ -129,78 +125,119 @@ class DeparturesRepositorySpec
...
     val result = await(repository.getDepartureIds(EORINumber("FR999")).value)
 
     result.right.get should be(None)
   }
 
   it should "return no departure ids when the db is empty" in {
-    await(repository.collection.drop().toFuture())
+    // the collection is empty at this point due to DefaultPlayMongoRepositorySupport
     val result = await(repository.getDepartureIds(EORINumber("FR999")).value)
     Thread.sleep(10000)
```
ardwalker, line 199: I'm being picky here, but sleeping for 10 seconds in a test might not be the best thing if it can be avoided

drnaylor-oc: I didn't spot that -- yeah, that needs to be looked at.

app/uk/gov/hmrc/transitmovements/repositories/DeparturesRepository.scala
drnaylor-oc: It'll return all messages for a given departure - however many that will be, and it'll only be IDs. There could be triple figures of messages on a departure in theory, but we're (thanfully) not returning all the contents of those, that could get rather large.

Messages are only purged when the departure itself is purged, which is 31 days after the last update to the departure (which will invariably be because a message was added to the departure).

it/uk/gov/hmrc/transitmovements/repositories/DeparturesRepositorySpec.scala
ardwalker: Yeah, I realise its outside your PR, but thought it was worth mentioning

derekspalmer: That's an error  that was left in.
Can be removed - the await catches it.

ardwalker (approved):

it/uk/gov/hmrc/transitmovements/repositories/DeparturesRepositorySpec.scala
drnaylor-oc: Yup, I'll ditch it now.

derekspalmer (approved):

PR #32, opened 2022-09-16
Made triggerId an optional parameter in updateMovement endpoint

liandra-giacomine:


RobertLaverick (commented):
Since we're no longer forcing the triggerId, it might be worth updating the MessageFactory tests to test all parameters are as expected rather than just the stream parsing.

PR #33, opened 2022-10-04
[CTCP-1354] renamed the collection to movements and added movementType to this collection

prashanthmamidi:
renamed Departure to Movement to support both arrival and departure
renamed the collection to movements
added movementType to this collection to flag whether something is an arrival or departure
Updated the departure queries to filter on departures


app/uk/gov/hmrc/transitmovements/models/MovementType.scala
```diff
@@ -0,0 +1,19 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.transitmovements.models
+
+case class MovementType(value: String) extends AnyVal
```
drnaylor-oc, line 19: As we only have two movement types, "arrival" and "departure", we should make this a sealed abstract class with two case objects, something like:

```scala
sealed abstract class MovementType(val value: String)

object MovementTypes {
    case object Arrival extends MovementType("arrival")
    case object Departure extends MovementType("departure")
}
```

and use these objects, rather than wrapping the strings in multiple places (that way, we're not repeating the magic string and if we need to refactor for whatever reason, it'll be easier to do - plus it avoids typos!).

prashanthmamidi: updated the code, please re-review

liandra-giacomine (commented):
@drnaylor-oc just want to confirm that the departure id will be different from the arrival id, rather than it being a single movement id? If it is the same id for both, updateMessages in the DeparturesRepository will have to be updated to also match with the movementType

drnaylor-oc:
They'll have the same form (16 character hex), as far the the DB is concerned they'll look the same. With that in mind, `DeparturesRepository` should become `MessagesRepository` as well, as it's one collection we'll be adding to.

app/uk/gov/hmrc/transitmovements/repositories/DeparturesRepository.scala
```diff
@@ -62,15 +63,28 @@ import scala.util.control.NonFatal
 @ImplementedBy(classOf[DeparturesRepositoryImpl])
 trait DeparturesRepository {
```
drnaylor-oc, line 64: This should become `MessagesRepository`

app/uk/gov/hmrc/transitmovements/repositories/DeparturesRepository.scala
```diff
@@ -62,15 +63,28 @@ import scala.util.control.NonFatal
...
 
   def getMessages(
     eoriNumber: EORINumber,
     departureId: DepartureId,
+    movementType: MovementType,
     received: Option[OffsetDateTime]
   ): EitherT[Future, MongoError, Option[NonEmptyList[MessageResponse]]]
-  def getDepartures(eoriNumber: EORINumber): EitherT[Future, MongoError, Option[NonEmptyList[DepartureWithoutMessages]]]
+
+  def getDepartures(eoriNumber: EORINumber, movementType: MovementType): EitherT[Future, MongoError, Option[NonEmptyList[DepartureWithoutMessages]]]
```
drnaylor-oc, line 87: This doesn't need the `MovementType` here, because it's always going to be a `Departure`. Instead, in the implementation of the message, we just hardcode the filter to use `MovementType.Departure`.

app/uk/gov/hmrc/transitmovements/repositories/DeparturesRepository.scala
```diff
@@ -62,15 +63,28 @@ import scala.util.control.NonFatal
 @ImplementedBy(classOf[DeparturesRepositoryImpl])
 trait DeparturesRepository {
   def insert(departure: Departure): EitherT[Future, MongoError, Unit]
-  def getDepartureWithoutMessages(eoriNumber: EORINumber, departureId: DepartureId): EitherT[Future, MongoError, Option[DepartureWithoutMessages]]
-  def getSingleMessage(eoriNumber: EORINumber, departureId: DepartureId, messageId: MessageId): EitherT[Future, MongoError, Option[MessageResponse]]
+
+  def getDepartureWithoutMessages(
+    eoriNumber: EORINumber,
+    departureId: DepartureId,
+    movementType: MovementType
```
drnaylor-oc, line 70: Same here - we'get getting a departure.

app/uk/gov/hmrc/transitmovements/repositories/DeparturesRepository.scala
```diff
@@ -119,9 +133,17 @@ class DeparturesRepositoryImpl @Inject() (
...
+  ): EitherT[Future, MongoError, Option[DepartureWithoutMessages]] = {
 
-    val selector   = mAnd(mEq("_id", departureId.value), mEq("enrollmentEORINumber", eoriNumber.value))
+    val selector = mAnd(
+      mEq("_id", departureId.value),
+      mEq("enrollmentEORINumber", eoriNumber.value),
+      mEq("movementType", movementType.value)
```
drnaylor-oc, line 145: The movement type will always be a `MovementType.Departure` here.

app/uk/gov/hmrc/transitmovements/repositories/DeparturesRepository.scala
```diff
@@ -199,9 +234,12 @@ class DeparturesRepositoryImpl @Inject() (
       )
     }
 
-  def getDepartures(eoriNumber: EORINumber): EitherT[Future, MongoError, Option[NonEmptyList[DepartureWithoutMessages]]] = {
-    val selector: Bson = mEq("enrollmentEORINumber", eoriNumber.value)
-    val projection     = DepartureWithoutMessages.projection
+  def getDepartures(eoriNumber: EORINumber, movementType: MovementType): EitherT[Future, MongoError, Option[NonEmptyList[DepartureWithoutMessages]]] = {
+    val selector: Bson = mAnd(
+      mEq("enrollmentEORINumber", eoriNumber.value),
+      mEq("movementType", movementType.value)
```
drnaylor-oc, line 240: The movement type will always be a `MovementType.Departure` here.

app/uk/gov/hmrc/transitmovements/services/DepartureFactory.scala
```diff
@@ -37,6 +38,7 @@ trait DepartureFactory {
 
   def create(
     eori: EORINumber,
+    movementType: MovementType,
```
drnaylor-oc, line 41: The movement type will always be a `MovementType.Departure` here - it's the `DepartureFactory`!

app/uk/gov/hmrc/transitmovements/services/DepartureFactory.scala
```diff
@@ -51,12 +53,14 @@ class DepartureFactoryImpl @Inject() (
...
```
drnaylor-oc, line 56: and here.

app/uk/gov/hmrc/transitmovements/services/DepartureFactory.scala
```diff
@@ -51,12 +53,14 @@ class DepartureFactoryImpl @Inject() (
...
     declarationData: DeclarationData,
     message: Message
   ): Departure =
     Departure(
       _id = DepartureId(ShortUUID.next(clock, random)),
       enrollmentEORINumber = eori,
+      movementType = movementType,
```
drnaylor-oc, line 63: and here

it/uk/gov/hmrc/transitmovements/it/generators/ModelGenerators.scala
```diff
@@ -99,11 +103,12 @@ trait ModelGenerators extends BaseGenerators {
     Arbitrary {
       for {
         id                      <- arbitrary[DepartureId]
+        movementType            <- arbitrary[MovementType]
```
drnaylor-oc, line 106: You might end up with a `Departure` that contains a `MovementType.Arrival`.

it/uk/gov/hmrc/transitmovements/repositories/DeparturesRepositorySpec.scala
```diff
@@ -195,25 +196,32 @@ class DeparturesRepositorySpec
...
+    val eoriGB       = arbitrary[EORINumber].sample.value
+    val eoriXI       = arbitrary[EORINumber].sample.value
+    val message      = arbitrary[Message].sample.value
+    val movementType = arbitrary[MovementType].sample.value
 
     val departureGB1 =
-      arbitrary[Departure].sample.value.copy(enrollmentEORINumber = eoriGB, created = instant, updated = instant, messages = NonEmptyList(message, List.empty))
+      arbitrary[Departure].sample.value.copy(
+        enrollmentEORINumber = eoriGB,
+        movementType = movementType,
```
drnaylor-oc, line 220: Just can be a `MovementType.Departure`

app/uk/gov/hmrc/transitmovements/models/Departure.scala
```diff
@@ -22,6 +22,7 @@ import java.time.OffsetDateTime
 
 case class Departure(
   _id: DepartureId,
+  movementType: MovementType,
```
drnaylor-oc, line 25: What happens if we do this?

```scala
case class Departure(_id: DepartureId, enrolementEORINumber: EORINumber, ...) {
    val movementType = MovementType.Departure
}
```

Will the various models include the movement type then? Would save having to add this to a departure each time.

(Alternatively, is this going to look the same as an arrival, barring the `movementReferenceNumber`? We might need to think on that, because if so, this could just be a `Movement`, in which case the variable movement type _is_ important.)

drnaylor-oc (commented):
Mostly just around how we're passing around `MovementType.Departure`.

**However**, thinking on this, we might actually be better off renaming `Departure` -> `Movement` and `DepartureId` -> `MovementId`, as they're going to have the same format, so it might be that these comments aren't really that relavent. I'm gonna think on this overnight -- @RobertLaverick and @liandra-giacomine might have a thought here too (so maybe we try to get a quick discussion about this going in the mobbing channel in the morning before you do anything to make sure you don't put effort in that'll get blown away).

app/uk/gov/hmrc/transitmovements/models/Departure.scala
liandra-giacomine: Maybe have a Movement class that the Departure and Arrival inherit from, where Movement would contain all of the shared variables

app/uk/gov/hmrc/transitmovements/repositories/DeparturesRepository.scala
prashanthmamidi: Yes, I was planning to rename this to MovementsRepository.  And to the other point @drnaylor-oc  mentioned, yes my next PR will include these changes to rename Departure to Movement and delete DepartureId and use MovementId instead.

drnaylor-oc: Yes, MovementsRepository, not MessagesRepository, silly me! I'm always getting those two confused when writing them down.

With regards to movement type -- this PR is just in a slightly weird state where we have Departure being supplied with a MovementType which is odd and makes me hesitant to merge as it stands because it it doesn't make sense. Similarly, you've moved the collection over to be "movements" and changed the model, but you haven't changed everything else around it.

I'd suggest making the rename changes now to make that consistent with the collection and model changes.

prashanthmamidi: yep, thought of doing this incrementally first, it's  better to rename theses changes as part of this PR  to be consistent.

app/uk/gov/hmrc/transitmovements/services/DepartureFactory.scala
prashanthmamidi: This will be renamed to MovementFactory, which supports both Arrival and Departure

app/uk/gov/hmrc/transitmovements/models/Departure.scala
prashanthmamidi: yes, Departure will be renamed to Movement

drnaylor-oc:
I don't see anything else wrong with this, I'm just gonna try the regression tests myself and see what happens (plus see if I can trigger the intermittent test you were getting).

app/uk/gov/hmrc/transitmovements/models/DepartureWithoutMessages.scala
```diff
@@ -23,9 +23,10 @@ import java.time.LocalDateTime
 import java.time.OffsetDateTime
 
 case class DepartureWithoutMessages(
-  _id: DepartureId,
+  _id: MovementId,
   enrollmentEORINumber: EORINumber,
   movementEORINumber: EORINumber,
+  movementType: MovementType,
```
drnaylor-oc, line 29: Remove this from the response model that goes back to the API layer. `MovementType` should be internal to this service only.

app/uk/gov/hmrc/transitmovements/models/responses/DeclarationResponse.scala
```diff
@@ -18,12 +18,12 @@ package uk.gov.hmrc.transitmovements.models.responses
...
 import uk.gov.hmrc.transitmovements.models.MessageId
+import uk.gov.hmrc.transitmovements.models.MovementId
 import uk.gov.hmrc.transitmovements.models.formats.PresentationFormats
 
 object DeclarationResponse extends PresentationFormats {
   implicit val declarationResponseFormat: OFormat[DeclarationResponse] = Json.format[DeclarationResponse]
 }
 
-case class DeclarationResponse(departureId: DepartureId, messageId: MessageId)
+case class DeclarationResponse(movementId: MovementId, messageId: MessageId)
```
drnaylor-oc, line 29: This should remain `departureId` because that's what the API expects for a declaration response.

drnaylor-oc (commented):
This has failed most of the regressions tests (if you haven't run them yet, run this service with the `CTC_TRADERS_API` profile in service manager, then clone https://github.com/hmrc/ctc-traders-api-tests and run `sbt test` on this repo), and you'll see a lot of failures.

The reason for the failures is because you've done one of two things:

* Added the message type to response models (e.g. DepartureWithoutMessages) -- this is an internal flag and doesn't need to go outside of this service
* You've renamed `departureId` to `movementId` in response models that get converted to Json, `common-transit-convention-traders` is still expecting this to be `departureId` (and this is correct as those endpoints only care for departures)

app/uk/gov/hmrc/transitmovements/models/responses/DeclarationResponse.scala
prashanthmamidi: ok, updated to use departureId

app/uk/gov/hmrc/transitmovements/models/DepartureWithoutMessages.scala
prashanthmamidi: yes, removed

app/uk/gov/hmrc/transitmovements/controllers/ArrivalsController.scala
```diff
@@ -0,0 +1,80 @@
...
+    with PresentationFormats {
+
+  def createArrival(eori: EORINumber): Action[Source[ByteString, _]] = Action.async(streamFromMemory) {
+    implicit request =>
+      withTemporaryFile {
+        (temporaryFile, source) =>
+          (for {
+            declarationData <- xmlParsingService.extractDeclarationData(source).asPresentation
+            fileSource = FileIO.fromPath(temporaryFile)
+            message <- messageFactory.create(MessageType.DeclarationData, declarationData.generationDate, None, fileSource).asPresentation
```
drnaylor-oc, line 70: The message type will be `ArrivalNotification`, not `DeclarationData`.

app/uk/gov/hmrc/transitmovements/models/ArrivalId.scala
```diff
@@ -0,0 +1,19 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.transitmovements.models
+
+case class ArrivalId(value: String) extends AnyVal
```
drnaylor-oc, line 19: Is this used now, I think we're just using `MovementId`? In that case, this should just be deleted.

app/uk/gov/hmrc/transitmovements/models/responses/ArrivalDeclarationResponse.scala
```diff
@@ -0,0 +1,29 @@
...
+
+package uk.gov.hmrc.transitmovements.models.responses
+
+import play.api.libs.json.Json
+import play.api.libs.json.OFormat
+import uk.gov.hmrc.transitmovements.models.ArrivalId
+import uk.gov.hmrc.transitmovements.models.MessageId
+import uk.gov.hmrc.transitmovements.models.formats.PresentationFormats
+
+object ArrivalDeclarationResponse extends PresentationFormats {
```
drnaylor-oc, line 25: `ArrivalNotificationResponse`

drnaylor-oc (commented):
Minor things - mostly that the initial arrivals message is an `ArrivalNotification` (which is a message type `IE007`, code `CC007C`)

app/uk/gov/hmrc/transitmovements/controllers/ArrivalsController.scala
```diff
@@ -0,0 +1,80 @@
...
+    with ConvertError
+    with TemporaryFiles
+    with PresentationFormats {
+
+  def createArrival(eori: EORINumber): Action[Source[ByteString, _]] = Action.async(streamFromMemory) {
+    implicit request =>
+      withTemporaryFile {
+        (temporaryFile, source) =>
+          (for {
+            declarationData <- xmlParsingService.extractDeclarationData(source).asPresentation
```
drnaylor-oc, line 68: Just thought about this, an arrival is going to pull out slightly different data, so this isn't going to work. This is what Sandeep was talking about in clarification.

We'll want to extract an `ArrivalData` or something like that - which extract the same things as the departure declaration, plus the MRN. Handily, [XmlParsers](https://github.com/hmrc/transit-movements/blob/main/app/uk/gov/hmrc/transitmovements/services/XmlParsers.scala#L35-L58) has extractors for each one, but only one of them has been parameterised to work based on the message type.

[You'll need to make one of these for Arrival Notifications](https://github.com/hmrc/transit-movements/blob/36d97c24044c623a9bc54c2ab779a148674282bd/app/uk/gov/hmrc/transitmovements/services/DeparturesXmlParsingService.scala#L65-L84), but it'll be a three item ZipWith that gets all the preparationDateTime, MRN and movement EORI.

app/uk/gov/hmrc/transitmovements/models/ArrivalId.scala
prashanthmamidi: yes, will delete DepartureId as well

app/uk/gov/hmrc/transitmovements/controllers/ArrivalsController.scala
prashanthmamidi: updated the logic arrival to extract the data

app/uk/gov/hmrc/transitmovements/controllers/ArrivalsController.scala
```diff
@@ -0,0 +1,79 @@
...
+) extends BackendController(cc)
+    with Logging
+    with StreamingParsers
+    with ConvertError
+    with TemporaryFiles
+    with PresentationFormats {
+
+  def createArrival(eori: EORINumber): Action[Source[ByteString, _]] = Action.async(streamFromMemory) {
+    implicit request =>
+      withTemporaryFile {
```
liandra-giacomine, line 64: Can we apply the .stream logic here? As implemented by @drnaylor-oc in common-transit-convention-traders/app/v2/controllers/stream/StreamingParsers.scala

app/uk/gov/hmrc/transitmovements/controllers/MovementsController.scala
```diff
@@ -71,7 +70,7 @@ class MovementsController @Inject() (
                   fileSource
                 )
                 .asPresentation
-              _ <- repo.updateMessages(DepartureId(movementId.value), message, messageData.mrn).asPresentation
+              _ <- repo.updateMessages(MovementId(movementId.value), message, messageData.mrn).asPresentation
```
liandra-giacomine, line 73: you can pass the movementId here, no need to create a new MovementId object and insert the value since the type is already as expected

app/uk/gov/hmrc/transitmovements/controllers/ArrivalsController.scala
drnaylor-oc: That's a separate ticket to do - not managed to get around to doing it on this repo yet. Feel free to leave it, we can handle that later. I don't want to stall this PR any longer.

app/uk/gov/hmrc/transitmovements/models/formats/PresentationFormats.scala
```diff
@@ -28,7 +28,7 @@ trait PresentationFormats extends CommonFormats {
   // as these use the OffsetDateTime times, these need to be here to avoid the macro expansion of
   // OffsetDateTime when used with Mongo.
   implicit val messageFormat: Format[Message]                                   = Json.format[Message]
-  implicit val departureFormat: Format[Departure]                               = Json.format[Departure]
+  implicit val departureFormat: Format[Movement]                                = Json.format[Movement]
```
drnaylor-oc, line 31: `movementFormat`

drnaylor-oc (commented):
Just one other small thing if you can just tweak this while handling Liandra's comment, then I think we'll be golden.

drnaylor-oc:
Just gonna give it a quick whirl locally!

drnaylor-oc (approved):
Yup, I'm happy. Merge when you're ready.

drnaylor-oc:
Ah... commit signing... right.

drnaylor-oc:
Superseded by #34

PR #34, opened 2022-10-17
[CTCP-1354] add arrival endpoint and use movement collection to support both arrivals and departures

prashanthmamidi:


drnaylor-oc (approved):
Just diffed this against the CTCP-1354 branch just for peace of mind, it's the same. So I'm happy.

PR #35, opened 2022-10-27
[CTCP-1486] added Get Arrival by EORI API

prashanthmamidi:


app/uk/gov/hmrc/transitmovements/models/formats/PresentationFormats.scala
```diff
@@ -30,6 +31,7 @@ trait PresentationFormats extends CommonFormats {
   implicit val messageFormat: Format[Message]                                   = Json.format[Message]
   implicit val movementFormat: Format[Movement]                                 = Json.format[Movement]
   implicit val departureWithoutMessagesFormat: Format[DepartureWithoutMessages] = Json.format[DepartureWithoutMessages]
```
liandra-giacomine, line 33: I don't think we still need departureWithoutMessagesFormat/DepartureWithoutMessages? If so, please delete

test/uk/gov/hmrc/transitmovements/controllers/ArrivalsControllerSpec.scala
```diff
@@ -323,4 +326,37 @@ class ArrivalsControllerSpec
     }
   }
 
+  "getArrivalsForEori" - {
+    val request = FakeRequest("GET", routes.ArrivalsController.getArrivalsForEori(eoriNumber).url)
+
+    "must return OK if arrivals were found" in {
+      val response = MovementWithoutMessages.fromMovement(movement)
+
+      when(mockRepository.getMovements(EORINumber(any()), any()))
```
liandra-giacomine, line 335: We should state the MovementType type used here, so for the getMovements calls, rather than any(), can we use eqTo(MovementType.Arrivals)?

test/uk/gov/hmrc/transitmovements/controllers/DeparturesControllerSpec.scala
```diff
@@ -427,9 +428,9 @@ class DeparturesControllerSpec
     val request = FakeRequest("GET", routes.DeparturesController.getDeparturesForEori(eoriNumber).url)
 
     "must return OK if departures were found" in {
-      val response = DepartureWithoutMessages.fromDeparture(movement)
+      val response = MovementWithoutMessages.fromMovement(movement)
 
-      when(mockRepository.getDepartures(EORINumber(any())))
+      when(mockRepository.getMovements(EORINumber(any()), any()))
```
liandra-giacomine, line 433: same thing for these getMovements calls, but with eqTo(MovementType.Departures)

app/uk/gov/hmrc/transitmovements/models/formats/PresentationFormats.scala
prashanthmamidi: yes, it should be deleted once this API - _**GET   /traders/:EORI/movements/departures/:movementId**_  is implemented

app/uk/gov/hmrc/transitmovements/controllers/ArrivalsController.scala
```diff
@@ -76,4 +77,17 @@ class ArrivalsController @Inject() (
       }.toResult
   }
 
+  def getArrivalsForEori(eoriNumber: EORINumber): Action[AnyContent] = Action.async {
+    repo
+      .getMovements(eoriNumber, MovementType.Arrival)
```
sonisn1988, line 82: @drnaylor-oc I think we should call this repo call via service call?  I never seen calling directly repo in controller. That is kind of tight binding between the controller and repository layer. This is just my observer. not sure if I am right were

drnaylor-oc: We're doing it elsewhere in this repo -- and it kind of makes sense because this service is _all_ about the Mongo DB and we don't do any additional processing beyond "turn it into a model and send it back in the response". 

Putting a service layer in the middle is a choice we could have made, and we looked at that early on -- but it would have added nothing for a higher level of complexity. So, for this service, I think it's okay.

app/uk/gov/hmrc/transitmovements/models/formats/PresentationFormats.scala
liandra-giacomine: Why not now? We're no longer using it right as it has been replaced with MovementWithoutMessages?

prashanthmamidi: thought of doing this earlier, as this is still being used for the other API _get single departure by departureid_  GET /traders/:EORI/movements/departures/:movementId

It would be good if we delete this as part of the below ticket 
https://jira.tools.tax.service.gov.uk/browse/CTCP-1489

liandra-giacomine: @prashanthmamidi Changing it to MovementWithoutMessages for that endpoint shouldn't affect anything since they are both the same model but with different names. I will write a note in the ticket so that the person implementing it for arrivals remembers to update the respective departures endpoint too.

drnaylor-oc (commented):
I think this is fine. I just want to draw your attention to what the pr-commenter said:

Looks like @Singleton is missing on [app/uk/gov/hmrc/transitmovements/repositories/MovementsRepository.scala](https://github.com/hmrc/transit-movements/blob/CTCP-1486/app/uk/gov/hmrc/transitmovements/repositories/MovementsRepository.scala#L100). Only one instance is preferable as indexes would be reinitalised - see https://github.com/hmrc/hmrc-mongo#playmongorepository

Can we do that? The other things I'm not so concerned about right this second, but adding `@Singleton` to the repository is something we really should do.

drnaylor-oc (approved):

PR #36, opened 2022-11-02
Get arrival messages by arrival id (transit-movements)

sonisn1988:


test/uk/gov/hmrc/transitmovements/controllers/ArrivalsControllerSpec.scala
```diff
@@ -360,4 +362,41 @@ class ArrivalsControllerSpec
     }
   }
 
+  "getArrivalMessages" - {
+
+    val request = FakeRequest("GET", routes.ArrivalsController.getArrivalMessages(eoriNumber, movementId).url)
+
+    "must return OK and a list of message ids" in {
```
drnaylor-oc, line 369: nitpick -- must return OK and a list of message ids if an arrival is found.

test/uk/gov/hmrc/transitmovements/controllers/ArrivalsControllerSpec.scala
```diff
@@ -360,4 +362,41 @@ class ArrivalsControllerSpec
...
+      val messageResponses = MessageResponse.fromMessageWithoutBody(movement.messages.head)
+
+      lazy val messageResponseList = NonEmptyList.one(messageResponses)
+
+      when(mockRepository.getMessages(EORINumber(any()), MovementId(any()), eqTo(MovementType.Arrival), eqTo(None)))
```
drnaylor-oc, line 374: We need to ensure that if the `receivedSince` parameter is passed down that it continues to be passed down to the repository.

We could use property based testing for this (using ScalaCheck and forall like you've hopefully seen in other tests - we should randomly generate the `OffsetDateTime` using a generator and use 

```
"must return OK and a list of message ids" in forall(Gen.option(dateTimeGenerator)) { 
    dateTimeMaybe => 
       ...
       when(mockRepository.getMessages(EORINumber(any()), MovementId(any()), eqTo(MovementType.Arrival), eqTo(dateTimeMaybe)))
      ...
```

Equally, the request should be generated for each test in this case, rather than global.

test/uk/gov/hmrc/transitmovements/controllers/ArrivalsControllerSpec.scala
sonisn1988: done

test/uk/gov/hmrc/transitmovements/controllers/ArrivalsControllerSpec.scala
sonisn1988: Done the changes as suggested above.

liandra-giacomine (approved):
LGTM

PR #49, opened 2023-01-12
CTCP-1921: Adding an empty movement for small messages

liandra-giacomine:
This has been implemented in respect to the following conclusions derived in a discussion between devs and QAs:

- We'll make messageId optional for the create a movement endpoint.

- We're going to check for the content-type header and if it's not present, assume it's a large message. We will need to ensure the API always sends the content-type for small messages.

- We're going to make the messages object in the Movement a List, there's no difference logically between List and Option[NonEmptyList] so there's no need to overcomplicate things. I have decided to use a Vector instead as it's more efficient than a List.

### DIFF
--- a/app/uk/gov/hmrc/transitmovements/controllers/MovementsController.scala
+++ b/app/uk/gov/hmrc/transitmovements/controllers/MovementsController.scala
@@ -70,10 +70,12 @@ class MovementsController @Inject() (
     with MessageTypeHeaderExtractor
     with PresentationFormats {
 
-  def createMovement(eori: EORINumber, movementType: MovementType): Action[Source[ByteString, _]] = movementType match {
-    case MovementType.Arrival   => createArrival(eori)
-    case MovementType.Departure => createDeparture(eori)
-  }
+  def createMovement(eori: EORINumber, movementType: MovementType): Action[Source[ByteString, _]] =
+    // Determine whether this is for a large or small message then route
+    movementType match {
+      case MovementType.Arrival   => createArrival(eori)
+      case MovementType.Departure => createDeparture(eori)
+    }
 
   private def createArrival(eori: EORINumber): Action[Source[ByteString, _]] = Action.streamWithAwait {
     awaitFileWrite => implicit request =>
@@ -84,7 +86,7 @@ class MovementsController @Inject() (
         message <- messageFactory.create(MessageType.ArrivalNotification, arrivalData.generationDate, received, None, request.body).asPresentation
         movement = movementFactory.createArrival(eori, MovementType.Arrival, arrivalData, message, received, received)
         _ <- repo.insert(movement).asPresentation
-      } yield MovementResponse(movement._id, movement.messages.head.id)).fold[Result](
+      } yield MovementResponse(movement._id, Some(movement.messages.get.head.id))).fold[Result](
         baseError => Status(baseError.code.statusCode)(Json.toJson(baseError)),
         response => Ok(Json.toJson(response))
       )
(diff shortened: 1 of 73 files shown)


drnaylor-oc:
I've been thinking about the idea for creating a message without actually having the message yet, why we would or wouldn't store the fact we're expecting a message. I realised why we might consider that.

Consider creating a new movement, and the trader submits a message. They decide not to subscribe to PPNS (for whatever reason) and decide to poll the endpoints. At first, that message will just not appear in the list, and this might confuse the trader. Do we need to think about that and how we do that?

Just a thought.

derekspalmer (approved):
LGTM

app/uk/gov/hmrc/transitmovements/controllers/MovementsController.scala
```diff
@@ -99,7 +106,20 @@ class MovementsController @Inject() (
...
         _ <- repo.insert(movement).asPresentation
-      } yield MovementResponse(movement._id, movement.messages.head.id)).fold[Result](
+      } yield MovementResponse(movement._id, Some(movement.messages.head.id))).fold[Result](
+        baseError => Status(baseError.code.statusCode)(Json.toJson(baseError)),
+        response => Ok(Json.toJson(response))
+      )
+  }
+
+  private def createEmptyMovement(eori: EORINumber, movementType: MovementType): Action[AnyContent] = Action.async(parse.anyContent) {
+    implicit requet =>
```
drnaylor-oc, line 116: `request`

app/uk/gov/hmrc/transitmovements/models/Movement.scala
```diff
@@ -16,17 +16,15 @@
...
   _id: MovementId,
   movementType: MovementType,
   enrollmentEORINumber: EORINumber,
-  movementEORINumber: EORINumber,
+  movementEORINumber: Option[EORINumber],
   movementReferenceNumber: Option[MovementReferenceNumber], // optional pending MRN allocation
   created: OffsetDateTime,
   updated: OffsetDateTime,
-  messages: NonEmptyList[Message]
+  messages: Vector[Message]
```
drnaylor-oc, line 29: Why a `Vector`? `Seq` should suffice -- we should let the runtime decide what's best unless we specifically need something from a more specialised type.

app/uk/gov/hmrc/transitmovements/repositories/MovementsRepository.scala
```diff
@@ -83,14 +82,14 @@ trait MovementsRepository {
     movementId: MovementId,
     movementType: MovementType,
     received: Option[OffsetDateTime]
-  ): EitherT[Future, MongoError, Option[NonEmptyList[MessageResponse]]]
+  ): EitherT[Future, MongoError, Vector[MessageResponse]]
```
drnaylor-oc, line 85: Vector or Seq? (I'll stop going on about it now, but you get the idea)

app/uk/gov/hmrc/transitmovements/services/MessageFactory.scala
```diff
@@ -32,7 +32,7 @@ import java.security.SecureRandom
 import java.time.Clock
 import java.time.OffsetDateTime
 import javax.inject.Inject
-import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.ExecutionContext
```
drnaylor-oc, line 35: Good spot.

drnaylor-oc (commented):
Nothing earth-shattering, but why `Vector` instead of `Seq`?

app/uk/gov/hmrc/transitmovements/controllers/MovementsController.scala
liandra-giacomine: Aha!

drnaylor-oc:
I just looked into the `Vector` stuff, I'm happy to leave it as is. I also thought `Seq` was backed by `Vector`, but it appears that's not the case:

```
Welcome to the Ammonite Repl 2.4.1 (Scala 2.13.7 Java 11.0.17)
@ val a = Seq("1", "a")
a: Seq[String] = List("1", "a")

@ a.getClass
res1: Class[T] = class scala.collection.immutable.$colon$colon

@ a.isInstanceOf[scala.collection.immutable.Vector[_]]
res2: Boolean = false
```

It's a `List` (that's what `::` is). I don't think it matters whatever we do, I was just curious as to why we used it here. I'm happy to keep it as is.

drnaylor-oc (approved):

PR #51, opened 2023-02-10
[CTCP-2408] return messageId for large message

prashanthmamidi:


app/uk/gov/hmrc/transitmovements/services/MessageFactory.scala
```diff
@@ -68,14 +74,29 @@ class MessageFactoryImpl @Inject() (
         Message(
           id = MessageId(ShortUUID.next(clock, random)),
           received = received,
-          generated = generationDate,
+          generated = Some(generationDate),
           messageType = messageType,
           triggerId = triggerId,
```
liandra-giacomine, line 79: when we create an empty message, we won't ever know the trigger id until later on, so can we remove it from the parameter list in the function and just set it to none in the object creation?

`def createEmptyMessage(
     messageType: MessageType,
     received: OffsetDateTime
   ): Message`

`triggerId = None,`

test/uk/gov/hmrc/transitmovements/services/MovementFactoryImplSpec.scala
```diff
@@ -84,16 +84,17 @@ class MovementFactoryImplSpec
   "createEmptyMovement" - {
     val sut = new MovementFactoryImpl(clock, random)
 
-    "will create a movement without a message, movementEORINumber and movementReferenceNumber" in forAll(
+    "will create a movement with message, movementEORINumber and movementReferenceNumber" in forAll(
```
liandra-giacomine, line 87: The description here indicates we have movementEORINumber and movementReferenceNumber, can we clarify that these won't be present

prashanthmamidi: yes, updated

app/uk/gov/hmrc/transitmovements/controllers/MovementsController.scala
```diff
@@ -119,11 +121,16 @@ class MovementsController @Inject() (
   private def createEmptyMovement(eori: EORINumber, movementType: MovementType): Action[AnyContent] = Action.async(parse.anyContent) {
     implicit request =>
       val received = OffsetDateTime.ofInstant(clock.instant, ZoneOffset.UTC)
-      val movement = movementFactory.createEmptyMovement(eori, movementType, received, received)
+      val messageType = movementType match {
+        case MovementType.Arrival   => MessageType.ArrivalNotification
+        case MovementType.Departure => MessageType.DeclarationData
+      }
+      val message  = messageFactory.createEmptyMessage(messageType, OffsetDateTime.ofInstant(clock.instant, ZoneOffset.UTC), MessageStatus.Pending)
```
liandra-giacomine, line 128: Better to pass received to the empty message so that it has the same creation time as the movement and doesn't cause issues when polling

`messageFactory.createEmptyMessage(messageType, received, MessageStatus.Pending)`

app/uk/gov/hmrc/transitmovements/services/MessageFactory.scala
```diff
@@ -80,6 +86,22 @@ class MessageFactoryImpl @Inject() (
...
+  ): Message =
+    Message(
+      id = MessageId(ShortUUID.next(clock, random)),
+      received = received,
+      generated = None,
+      messageType = messageType,
+      triggerId = None,
+      url = None,
+      body = None,
+      status = status
```
liandra-giacomine, line 102: status here will always be pending, so rather than passing it as a parameter, just set it in the object creation

`status = Pending`

app/uk/gov/hmrc/transitmovements/controllers/MovementsController.scala
```diff
@@ -119,11 +121,16 @@ class MovementsController @Inject() (
   private def createEmptyMovement(eori: EORINumber, movementType: MovementType): Action[AnyContent] = Action.async(parse.anyContent) {
     implicit request =>
       val received = OffsetDateTime.ofInstant(clock.instant, ZoneOffset.UTC)
-      val movement = movementFactory.createEmptyMovement(eori, movementType, received, received)
+      val messageType = movementType match {
```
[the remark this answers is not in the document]
drnaylor-oc: It really depends on what we do in the callback, but if we have "one callback fits all" from Upscan, then the message type will be set there anyway.

Alternatively, we keep it, and when in the callback, we use that as a "this is what we're expecting, beyond a departure message". Either way could work.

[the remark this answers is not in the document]
sonisn1988: but if we are storing initially then later we need to update it in same cases. Like if traders pass some other arrival or departure message type apart from IE007 and IE015. So better to insert messageType at call back endpoint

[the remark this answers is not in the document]
liandra-giacomine: No, if they submit something which isn't IE007 or IE015 when creating an empty movement then we won't accept it and store it

liandra-giacomine (approved):

app/uk/gov/hmrc/transitmovements/controllers/MovementsController.scala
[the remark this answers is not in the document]
sonisn1988: Might be my understanding is wrong. Just wondering how we are going to handle attachMessage (existing endpoint) for large message? if we are going to accept only IE007 or IE015 messageType with callback? is going to be different callback for attach Large message?

PR #53, opened 2023-02-24
Attach new large message from ERMIS to existing movement (transit-mov…

sonisn1988:
…ements)


app/uk/gov/hmrc/transitmovements/connectors/ObjectStoreConnector.scala
```diff
@@ -0,0 +1,56 @@
...
+
+  override def getObjectStoreFile(objectStoreURL: String)(implicit
+    ec: ExecutionContext,
+    hc: HeaderCarrier
+  ): Future[Source[ByteString, _]] =
+    client
+      .getObject[Source[ByteString, NotUsed]](Path.File(objectStoreURL), "common-transit-conversion-traders")
+      .flatMap {
+        case Some(source) => Future.successful(source.content)
+        case _            => Future.failed((new RuntimeException("File not found")))
```
liandra-giacomine, line 53: It's better to return an Option[Source[ByteString,_]] than a failed future with an exception

app/uk/gov/hmrc/transitmovements/controllers/errors/ConvertError.scala
```diff
@@ -67,6 +68,16 @@ trait ConvertError {
     }
   }
 
+  implicit val objectStoreErrorConverter = new Converter[ObjectStoreError] {
+
+    import uk.gov.hmrc.transitmovements.services.errors.ObjectStoreError._
+
+    def convert(objectStoreError: ObjectStoreError): PresentationError = objectStoreError match {
+      case FileNotFound(fileLocation) => PresentationError.notFoundError(s"file not found at location: $fileLocation")
```
liandra-giacomine, line 76: Would this be an internal server error instead?

app/uk/gov/hmrc/transitmovements/controllers/MovementsController.scala
```diff
@@ -136,7 +129,37 @@ class MovementsController @Inject() (
...
+      case Some(_) => updateMovementSmallMessage(movementId, triggerId)
+      case None    => updateMovementLargeMessage(movementId, triggerId)
+    }
+
+  def updateMovementLargeMessage(movementId: MovementId, triggerId: Option[MessageId] = None): Action[AnyContent] = Action.async(parse.anyContent) {
+    implicit request =>
+      (for {
+        messageType    <- extract(request.headers).asPresentation
+        objectStoreURL <- extractObjectStoreURI(request.headers).asPresentation
+        objectStoreFilePath = if (objectStoreURL.contains("/")) objectStoreURL.split("/", 2).apply(1) else objectStoreURL
```
liandra-giacomine, line 143: can we do this in the objectStore object and returned as a value instead?

app/uk/gov/hmrc/transitmovements/controllers/ObjectStoreURIHeaderExtractor.scala
```diff
@@ -0,0 +1,36 @@
...
+import play.api.mvc.Headers
+import uk.gov.hmrc.transitmovements.config.Constants
+import uk.gov.hmrc.transitmovements.controllers.errors.HeaderExtractError
+import uk.gov.hmrc.transitmovements.controllers.errors.HeaderExtractError.NoHeaderFound
+
+import scala.concurrent.Future
+
+trait ObjectStoreURIHeaderExtractor {
+
+  def extractObjectStoreURI(headers: Headers): EitherT[Future, HeaderExtractError, String] =
```
liandra-giacomine, line 29: Can we create a case class for the ObjectStoreURI instead of a string

app/uk/gov/hmrc/transitmovements/controllers/MovementsController.scala
```diff
@@ -136,7 +129,37 @@ class MovementsController @Inject() (
...
+        sourceFile  <- objectStoreService.getObjectStoreFile(objectStoreFilePath).asPresentation
+        messageData <- messagesXmlParsingService.extractMessageData(sourceFile, messageType).asPresentation
+        received = OffsetDateTime.ofInstant(clock.instant, ZoneOffset.UTC)
+        message = messageFactory
+          .createLargeMessage(
+            messageType,
+            messageData.generationDate,
+            received,
+            triggerId,
+            new URI(objectStoreURL)
```
liandra-giacomine, line 153: I'd rather this was stored as our own case class rather than URI

app/uk/gov/hmrc/transitmovements/controllers/MovementsController.scala
```diff
@@ -136,7 +129,37 @@ class MovementsController @Inject() (
...
+          )
+        _ <- repo.updateMessages(movementId, message, messageData.mrn, received).asPresentation
+      } yield message.id).fold[Result](
+        baseError => Status(baseError.code.statusCode)(Json.toJson(baseError)),
+        id => Ok(Json.toJson(UpdateMovementResponse(id)))
+      )
+  }
+
+  private def updateMovementSmallMessage(movementId: MovementId, triggerId: Option[MessageId] = None): Action[Source[ByteString, _]] =
```
liandra-giacomine, line 162: Might be good to create a Small messages and Large messages controller, but not required in this PR

app/uk/gov/hmrc/transitmovements/connectors/ObjectStoreConnector.scala
```diff
@@ -0,0 +1,56 @@
...
+trait ObjectStoreConnector {
+
+  def getObjectStoreFile(objectStoreURL: String)(implicit
+    ec: ExecutionContext,
+    hc: HeaderCarrier
+  ): Future[Source[ByteString, _]]
+
+}
+
+class ObjectStoreConnectorImpl @Inject() (client: PlayObjectStoreClient) extends ObjectStoreConnector {
```
drnaylor-oc, line 43: I don't think this should exist, and we should use the client in the service.

Connectors are used for direct outbound HTTP calls. The client does that for us and has its own connector -- so it's appropriate to call this in the service, in the same way we do explicit audits.

app/uk/gov/hmrc/transitmovements/controllers/MovementsController.scala
drnaylor-oc: I agree, it's the service's job to do this, not the presentation layer's.

app/uk/gov/hmrc/transitmovements/controllers/errors/ConvertError.scala
drnaylor-oc: Bad request. The header provided by the calling service is bad, because it doesn't point to a file. Not this service's fault.

it/uk/gov/hmrc/transitmovements/connectors/BaseUnitSpec.scala
```diff
@@ -0,0 +1,31 @@
...
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.ExecutionContext.global
+
+trait BaseUnitSpec {
+  implicit protected val ec: ExecutionContext         = global
+  implicit protected val system: ActorSystem          = ActorSystem()
+  implicit protected val materializer: Materializer   = Materializer(system)
+  implicit protected val headerCarrier: HeaderCarrier = HeaderCarrier()
+}
```
drnaylor-oc, line 31: I'm not in principle opposed to this, but keep it in line with other repos. Find an instance of `TestActorSystem` and use that instead.

I don't think the header carrier should be provided like this in a trait. The form of it should be clear to the test because it influences the test.

it/uk/gov/hmrc/transitmovements/connectors/ObjectStoreConnectorSpec.scala
```diff
@@ -0,0 +1,73 @@
...
+import uk.gov.hmrc.objectstore.client.play.PlayObjectStoreClient
+import uk.gov.hmrc.transitmovements.connectors.BaseUnitSpec
+import uk.gov.hmrc.transitmovements.connectors.ObjectStoreConnectorImpl
+
+import java.time.Instant
+import java.util.UUID.randomUUID
+import scala.concurrent.Future
+import scala.util.Try
+
+class ObjectStoreConnectorSpec extends AnyWordSpec with Matchers with MockitoSugar with ScalaFutures with BaseUnitSpec {
```
drnaylor-oc, line 40: I realise that this is a connector, and I realise I've told you to take it out, but these aren't integration tests anyway -- they are unit tests with mocked interfaces. These would really be unit tests.

it/uk/gov/hmrc/transitmovements/connectors/ObjectStoreConnectorSpec.scala
```diff
@@ -0,0 +1,73 @@
...
+    }
+  }
+
+  class ObjectStoreConnectorFixture {
+    val filename                                                                    = "movements/movementId/abc.xml"
+    val objectStoreClient                                                           = mock[PlayObjectStoreClient]
+    val metadata                                                                    = ObjectMetadata("", 0, Md5Hash(""), Instant.now(), Map.empty[String, String])
+    val content                                                                     = "content"
+    def content(value: String = randomUUID().toString): Source[ByteString, NotUsed] = Source.single(ByteString(value))
+    val o                                                                           = Option[Object[Source[ByteString, NotUsed]]](Object.apply(File(filename), Source.single(ByteString(content)), metadata))
```
drnaylor-oc, line 69: Something I've been pulled up for many times, don't use single letter field names. (They're generally fine in lambdas, sometimes in local declarations, but not in fields)

it/uk/gov/hmrc/transitmovements/connectors/ObjectStoreConnectorSpec.scala
```diff
@@ -0,0 +1,73 @@
...
+
+import java.time.Instant
+import java.util.UUID.randomUUID
+import scala.concurrent.Future
+import scala.util.Try
+
+class ObjectStoreConnectorSpec extends AnyWordSpec with Matchers with MockitoSugar with ScalaFutures with BaseUnitSpec {
+
+  "ObjectStoreConnector" should {
+    "return the contents of a file" in new ObjectStoreConnectorFixture {
```
drnaylor-oc, line 43: I'm generally not a fan of having inner classes per test, because you're instantiating everything each time, and you've hidden the setup for the test, which I'd argue is important to have explicitly stated so that the tests can be used as a reference for behaviour.

In fact, the mock you use in your fixture is used in this test, and ignored in the second test -- so what's the point of setting all that up? Additionally, your mocks just use `any()`, so you may as well just let a generator generate the filenames, and such -- defeats the point of the fixture really, particularly in the second test.

If you have to set up an environment for a bunch of tests and you need to be able to tear it down and refresh it, and `before/after` methods won't do, then some class/trait does work quite nicely. 

I don't think it works here.

app/uk/gov/hmrc/transitmovements/services/ObjectStoreService.scala
```diff
@@ -0,0 +1,63 @@
...
+
+}
+
+@Singleton
+class ObjectStoreServiceImpl @Inject() (client: PlayObjectStoreClient) extends ObjectStoreService {
+
+  override def getObjectStoreFile(
+    objectStoreURI: ObjectStoreURI
+  )(implicit ec: ExecutionContext, hc: HeaderCarrier): EitherT[Future, ObjectStoreError, Source[ByteString, _]] = {
+    val objectStoreFilePath = if (objectStoreURI.value.contains("/")) objectStoreURI.value.split("/", 2).apply(1) else objectStoreURI.value
```
drnaylor-oc, line 50: This line feels incredibly opaque to me. What is this line doing?

test/uk/gov/hmrc/transitmovements/services/ObjectStoreServiceSpec.scala
```diff
@@ -0,0 +1,110 @@
...
+      val metadata    = ObjectMetadata("", 0, Md5Hash(""), Instant.now(), Map.empty[String, String])
+      val content     = "content"
+      val fileContent = Option[Object[Source[ByteString, NotUsed]]](Object.apply(File(filePath), Source.single(ByteString(content)), metadata))
+      when(mockObjectStoreClient.getObject[Source[ByteString, NotUsed]](any[File](), any())(any(), any())).thenReturn(Future.successful(fileContent))
+      val service = new ObjectStoreServiceImpl(mockObjectStoreClient)
+      val result  = service.getObjectStoreFile(ObjectStoreURI(filePath))
+      whenReady(result.value) {
+        r =>
+          r.isRight mustBe true
+          r.toOption.get mustBe fileContent.get.content
```
drnaylor-oc, line 82: Can this be shortened to `r mustBe Right(content)`?

test/uk/gov/hmrc/transitmovements/services/ObjectStoreServiceSpec.scala
```diff
@@ -0,0 +1,110 @@
...
+
+    "should return an error when the file is not found on path" in {
+      when(mockObjectStoreClient.getObject(any[File](), any())(any(), any())).thenReturn(Future.successful(None))
+      val service = new ObjectStoreServiceImpl(mockObjectStoreClient)
+      val result  = service.getObjectStoreFile(ObjectStoreURI("abc/movement/abc.xml"))
+      whenReady(result.value) {
+        r =>
+          r.isLeft mustBe true
+          r.left.getOrElse(ObjectStoreError.FileNotFound) mustBe a[ObjectStoreError.FileNotFound]
+      }
```
drnaylor-oc, line 95: Maybe use pattern matching instead? Something like...

```scala
whenReady(result.value) {
    case Left(_: ObjectStoreError.FileNotFound) => succeed
    case x => fail(s"Expected Left(ObjectStoreError.FileNotFound), instead got $x")
}
```

Just looking for something that's more readable (I realise this might have been taken from previous tests (probably mine), but we should look to improve as we learn more about what we can do with Scala and such.)

conf/application.conf
```diff
@@ -120,6 +121,12 @@ microservice {
       host = localhost
       port = 8500
     }
+    object-store {
+      host = localhost
+      port = 8464
+    }
```
drnaylor-oc, line 127: Make sure you define the correct values in `app-config-base`

app/uk/gov/hmrc/transitmovements/services/ObjectStoreService.scala
sonisn1988: We will getting objectStoreURI like common-transit-convention-traders/movements/:movementID ..... So with that we need to remove common-transit-convention-traders service part in order to get the file. I have tested this with Object store stub and looks good. When I am including common-transit-convention-traders with get request I am getting 404 File not Found. For reference please check: https://github.com/hmrc/object-store-client#get-object

test/uk/gov/hmrc/transitmovements/services/ObjectStoreServiceSpec.scala
sonisn1988: I tried but I am getting  Right(Source(SourceShape(single.out(1854403227)))) was not equal to Right("content") . So I used r.toOption.get mustBe fileContent.get.content

app/uk/gov/hmrc/transitmovements/services/ObjectStoreService.scala
drnaylor-oc: Right -- but that's not obvious to someone reading the code itself. This line is actually making an assumption about the URI provided -- and a valid one at that, but by just reading the code, without prior knowledge I don't know what it is. If/when you look at it in six months, or if Live Services touch it, that knowledge may not be there.

I'm still nervous about this because of what you're doing if the URI doesn't contain a `/`. If that happens, we know things have just gone plain wrong anyway. Additionally, we know that the URI must **always** be for `common-transit-convention-traders`, right?

So, I propose something a little tighter, something like this on `ObjectStoreURI`:

```
object  ObjectStoreURI {

    // The URI consists of the service name in the first part of the path, followed
    // by the location of the object in the context of that service. As this service
    // targets common-transit-convention-traders' objects exclusively, we ensure
    // the URI is targeting that context. This regex ensures that this is the case.
    val expectedUriPattern = "^common-transit-convention-traders/(.+)$".r

}

case class ObjectStoreURI(value: String) extends AnyVal {

    def path: Option[String] = value match {
        case ObjectStoreURI.expectedUriPattern(path) => Some(path)
        case _ => None
    }
}
```

Of course, this logic can be in the service too. Not bothered where it ends up. Then, you can handle the path you've just extracted easily.

sonisn1988: Done the changes as suggested above

app/uk/gov/hmrc/transitmovements/services/ObjectStoreService.scala
```diff
@@ -0,0 +1,64 @@
...
+@Singleton
+class ObjectStoreServiceImpl @Inject() (client: PlayObjectStoreClient) extends ObjectStoreService {
+
+  override def getObjectStoreFile(
+    objectStoreURI: ObjectStoreURI
+  )(implicit ec: ExecutionContext, hc: HeaderCarrier): EitherT[Future, ObjectStoreError, Source[ByteString, _]] =
+    EitherT(
+      client
+        .getObject[Source[ByteString, NotUsed]](
+          Path.File(objectStoreURI.path.map(_.mkString).getOrElse(objectStoreURI.value)),
```
drnaylor-oc, line 53: If I send down `transit-movements-router/movements/abcdef0123456789/message/abcdef-20230701-120000.xml`, that's a mistake, right? That lets this through. Returning a `None` is an error condition so we should be treating it as such, right?

sonisn1988: ok. so do you want me to use like objectStoreURI.path.map(_.mkString) only? so sending transit-movements-router/movements/abcdef0123456789/message/abcdef-20230701-120000.xml will return None here.

drnaylor-oc: Nope. That's correct.

drnaylor-oc: This line still doesn't reflect what I asked, and it doesn't really reflect what you said. This line is calling `object-store` when we **know** the result will be wrong, which we should not be doing. If we can't extract the path because it's of the wrong form, we should fail fast and NOT call object-store just to get an error code.

Whatever is in the option returned back from `objectStoreURI.path` should be the ONLY value that we consider here. I really don't know why we have the `getOrElse` here just using the URI, because by out specification, that's always going to be wrong, so why then reach out to object-store to determine that, just to then return the error you could short circuit?

Perhaps consider breaking this function up a little and putting it behind a for comprehension if you're worried about trying to make this totally functional:

```scala
override def getObjectStoreFile(
    objectStoreURI: ObjectStoreURI
  )(implicit ec: ExecutionContext, hc: HeaderCarrier): EitherT[Future, ObjectStoreError, Source[ByteString, _]] =
    for {
        path   <- getPath(objectStoreURI)
        result <- getObjectFromObjectStore(path)
    } yield result

private def getPath(objectStoreUri: ObjectStoreURI)(implicit ec: ExecutionContext): EitherT[Future, ObjectStoreError, Source[ByteString, _]] =
    EitherT(Future.successful(objectStoreUri.path.toRight(ObjectStoreError.InvalidURI)))

private def getObjectFromObjectStore(path: String)(implicit ec: ExecutionContext, hc: HeaderCarrier): EitherT[Future, ObjectStoreError, Source[ByteString, _]] = ???
```

Also, to note:

* No need to call `mkString` on a string, which you do in `objectStoreURI.path.map(_.mkString)`
* You repeated `objectStoreURI.path.map(_.mkString).getOrElse(objectStoreURI.value)` below on line 58 -- be careful not to repeat yourself in these cases.

app/uk/gov/hmrc/transitmovements/services/errors/ObjectStoreError.scala
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.transitmovements.services.errors
+
+sealed abstract class ObjectStoreError extends Throwable
+
+object ObjectStoreError {
+  case class UnexpectedError(caughtException: Option[Throwable] = None) extends ObjectStoreError
+  case class FileNotFound(fileLocation: String)                         extends ObjectStoreError
```
drnaylor-oc, line 23: You'll want another case here for an invalid URI, and in the converter that will convert to a bad request.

app/uk/gov/hmrc/transitmovements/services/ObjectStoreService.scala
sonisn1988: ohk I got your point. In that case I think better to handle in it in ObjectStoreURIHeaderExtractor  instead of doing it in Service so that we can catch it while validating ObjectStoreURIHeader. Please suggest if below one is looks good
trait ObjectStoreURIHeaderExtractor {

  def extractObjectStoreURI(headers: Headers): EitherT[Future, HeaderExtractError, ObjectStoreURI] =
    EitherT {
      headers.get(Constants.ObjectStoreURI) match {
        case None => Future.successful(Left(NoHeaderFound("Missing X-Object-Store-Uri header value")))
        case Some(headerValue) =>
          ObjectStoreURI(headerValue).path match {
            case None                 => Future.successful(Left(InvalidObjectStoreURI(s"Invalid X-Object-Store-Uri header value: $headerValue")))
            case Some(objectStoreURI) => Future.successful(Right(ObjectStoreURI(objectStoreURI)))
          }
      }
    }

drnaylor-oc: If you like, but I'd rather not have nested pattern matching if we can help it.

app/uk/gov/hmrc/transitmovements/controllers/ObjectStoreURIHeaderExtractor.scala
```diff
@@ -0,0 +1,42 @@
...
+  def extractObjectStoreURI(headers: Headers): EitherT[Future, HeaderExtractError, ObjectStoreURI] =
+    EitherT {
+      headers.get(Constants.ObjectStoreURI) match {
+        case None => Future.successful(Left(NoHeaderFound("Missing X-Object-Store-Uri header value")))
+        case Some(headerValue) =>
+          ObjectStoreURI(headerValue).path match {
+            case None                 => Future.successful(Left(InvalidObjectStoreURI(s"Invalid X-Object-Store-Uri header value: $headerValue")))
+            case Some(objectStoreURI) => Future.successful(Right(ObjectStoreURI(objectStoreURI)))
+          }
+      }
```
drnaylor-oc, line 40: You've created another problem here by doing this now, `ObjectStoreURI` now represents both the full URI, and the fragment that does not contain `common-transit-convention-traders`, and that's bad because `ObjectStoreURI` was created as a marker to indicate a specific type of value... now you have two types of values under the same marker.

I also suggested what I did before because you were passing this whole URI to a service, which could then get the information from the value. You're now using `ObjectStoreURI` to store the computed value, so `path` shouldn't exist on it.

If you're going to go down this route, remove `path` from `ObjectStoreURI` and do it in here instead. Something like the following will possibly work:

```scala
object ObjectStoreURIHeaderExtractor {

  // The URI consists of the service name in the first part of the path, followed
  // by the location of the object in the context of that service. As this service
  // targets common-transit-convention-traders' objects exclusively, we ensure
  // the URI is targeting that context. This regex ensures that this is the case.
  val expectedUriPattern: Regex = "^common-transit-convention-traders/(.+)$".r
}
trait ObjectStoreURIHeaderExtractor {

  def extractObjectStoreURI(headers: Headers): EitherT[Future, PresentationError, ObjectStoreURI] =
    EitherT(
      Future.successful(
        for {
          headerValue    <- headers.get(Constants.ObjectStoreURI).toRight(PresentationError.badRequestError("Missing X-Object-Store-Uri header value"))
          objectStoreUri <- expectedUriPattern.findFirstMatchIn(headerValue).map(_.group(1)).toRight(PresentationError.badRequestError(s"X-Object-Store-Uri header value does not start with common-transit-convention-traders/ (got $headerValue)"))
        } yield ObjectStoreURI(objectStoreUri)
      )
    )
}
```

Note the following:

* `ObjectStoreURI` is a bad name now, I suggest something like `ObjectStoreResourceLocation` (because this object will no longer hold a full URI).
* This trait is intended to be mixed into a presentation layer object, because you're extracting from a header. In this case, just use `PresentationError` objects directly. Liandra self typed her presentation layer trait to `BaseController`, you could do that too. 
* You can split the two lines out into their own functions if you want.
* Check what I've written is correct and don't take it for granted, it's not completely tested.

drnaylor-oc (commented):
I don't have further comments, but as a bit of my code has been incorporated to this I would like someone else to review this.

PR #55, opened 2023-03-02
[CTCP-2017] a new endpoint to update message

prashanthmamidi:


app/uk/gov/hmrc/transitmovements/controllers/MovementsController.scala
```diff
@@ -135,6 +142,62 @@ class MovementsController @Inject() (
...
+          received = OffsetDateTime.ofInstant(clock.instant, ZoneOffset.UTC)
+          message = messageFactory.updateLargeMessage(
+            messageType,
+            Some(messageData.generationDate),
+            received,
+            messageId,
+            largeMessageMetadata.objectStoreURI,
+            status
+          )
+          _ <- repo.updateMessages(movementId, message, messageData.mrn, received).asPresentation
```
liandra-giacomine, line 176: We need a new function to update the array field in the message. This function adds a message to the movement rather than updates fields in a message in a movement. It should be renamed to something like addMessage and a new function created for the updating.

app/uk/gov/hmrc/transitmovements/controllers/MovementsController.scala
```diff
@@ -135,6 +142,62 @@ class MovementsController @Inject() (
...
+    }
+
+  private def handleUpdateMessage(largeMessageMetadata: LargeMessageMetadata, request: Request[JsValue], movementId: MovementId, messageId: MessageId)(implicit
+    hc: HeaderCarrier
+  ) =
+    largeMessageMetadata match {
+      case LargeMessageMetadata(Some(_), status) =>
+        (for {
+          messageType <- extract(request.headers).asPresentation
+          sourceFile  <- objectStoreService.getObjectStoreFile(largeMessageMetadata.objectStoreURI.get).asPresentation
```
liandra-giacomine, line 165: We don't need to extract the message data and get the file from object-store, so this can be just one for comprehension. Here all we care about is updating a message id which matches the message id in the request, held inside a movement whose _id matches the movement id sent in the request. We then update the url field with the object-store url if present, and update the status field with the status sent. We're not creating a new message, we're updating fields in an existing message.

liandra-giacomine (changes requested):
This isn't being implemented correctly at the moment.

app/uk/gov/hmrc/transitmovements/controllers/MovementsController.scala
liandra-giacomine: The metadata extraction would already have been done at this point

prashanthmamidi: updated

app/uk/gov/hmrc/transitmovements/controllers/MovementsController.scala
prashanthmamidi: added a new method for updateMessage

app/uk/gov/hmrc/transitmovements/controllers/MovementsController.scala
```diff
@@ -135,6 +142,37 @@ class MovementsController @Inject() (
       case None    => updateMovementLargeMessage(movementId, triggerId)
     }
 
+  def updateMessage(movementId: MovementId, messageId: MessageId) =
+    Action.async(parse.json) {
+      implicit request =>
+        mapRequest(request.body)
+          .flatMap {
+            largeMessageMetadata => handleUpdateMessage(largeMessageMetadata, movementId, messageId)
```
liandra-giacomine, line 150: largeMessageMetadata, could be a small message so can we rename it to messageMetadata instead?

app/uk/gov/hmrc/transitmovements/controllers/MovementsController.scala
```diff
@@ -135,6 +142,37 @@ class MovementsController @Inject() (
       case None    => updateMovementLargeMessage(movementId, triggerId)
     }
 
+  def updateMessage(movementId: MovementId, messageId: MessageId) =
+    Action.async(parse.json) {
+      implicit request =>
+        mapRequest(request.body)
```
liandra-giacomine, line 148: we can move handleUpdateMessage into this function since it's simple now.

(for {
 messageMetadata <- mapRequest(request.body)
  update <- repo.updateMessage(movementId, messageId, largeMessageMetadata, OffsetDateTime.ofInstant(clock.instant, ZoneOffset.UTC)).asPresentation
} yield update).fold[Result](
             presentationError => Status(presentationError.code.statusCode)(Json.toJson(presentationError)),
             _ => Ok
           )

app/uk/gov/hmrc/transitmovements/models/formats/MongoFormats.scala
```diff
@@ -53,6 +54,7 @@ trait MongoFormats extends CommonFormats with MongoBinaryFormats.Implicits with
   implicit val movementFormat: Format[Movement]                               = Json.format[Movement]
   implicit val movementWithoutMessagesFormat: Format[MovementWithoutMessages] = Json.format[MovementWithoutMessages]
   implicit val messageResponseFormat: Format[MessageResponse]                 = Json.format[MessageResponse]
+  //implicit val messageStatusFormat: Format[MessageStatus]                     = Json.format[MessageStatus]
```
liandra-giacomine: This is commented out, did it not work when you had the codec in and used val setStatus = mSet("messages.$[element].status", message.status)? If not then can we remove the commented out code?

it/uk/gov/hmrc/transitmovements/repositories/MovementsRepositorySpec.scala
```diff
@@ -536,4 +538,75 @@ class MovementsRepositorySpec
...
+    }
+
+    movement.updated shouldEqual receivedInstant
+    movement.messages.length should be(1)
+    movement.messages.head.status shouldBe Some(message2.status)
+    movement.messages.head.url.value.toString shouldBe message2.objectStoreURI.get.value
+
+  }
+
+  "updateMessage" should "update an existing message with status only" in {
```
liandra-giacomine, line 578: we need a test for DocumentNotFound

liandra-giacomine (commented):
A few nitpicks but almost good to go

app/uk/gov/hmrc/transitmovements/models/formats/MongoFormats.scala
prashanthmamidi: yes, removed this

liandra-giacomine (approved):

PR #58, opened 2023-03-16
CTCP-2488 configure small message changes and CTCP-2552 added tests

akila-sankar:


app/uk/gov/hmrc/transitmovements/utils/SmallMessageLimit.scala
```diff
@@ -0,0 +1,33 @@
...
+import javax.inject.Singleton
+
+@Singleton
+class SmallMessageLimit @Inject() (config: AppConfig) {
+  private lazy val limit = config.smallMessageSizeLimit
+
+  def checkContentSize(size: Long): Boolean = size match {
+    case (x) if x > limit => true
+    case (_)              => false
+  }
```
drnaylor-oc, line 31: This is a bit overkill for what you're doing here - it boils down to

```scala
def checkContentSize(size: Long): Boolean = size > limit
```

Does this warrant a separate class like this? I imagine you've done this because of the injection of `AppConfig`. At most, I'd probably make this a `trait` in the `controllers` package, possibly something like

```scala
trait SmallMessageLimit {

    def config: AppConfig // supply this in the controller constructor

    private lazy val limit = config.smallMessageSizeLimit

    def checkContentSize(size: Long): Boolean = size > limit
}
```

which would still be testable by doing

```
object Harness extends SmallMessageLimit {
    override val config: AppConfig = ??? // dummy impl here
}

Harness.checkContentSize(123)
```

That said, if you want to keep this as a separate class, then because you're injecting things, make this the `SmallMessageLimitService` and put it in the `services` package.

app/uk/gov/hmrc/transitmovements/controllers/MovementsController.scala
```diff
@@ -84,36 +91,93 @@ class MovementsController @Inject() (
...
-          .create(MessageType.ArrivalNotification, arrivalData.generationDate, received, None, request.body, MessageStatus.Processing)
-          .asPresentation
-        movement = movementFactory.createArrival(eori, MovementType.Arrival, arrivalData, message, received, received)
-        _ <- repo.insert(movement).asPresentation
-      } yield MovementResponse(movement._id, Some(movement.messages.head.id))).fold[Result](
-        baseError => Status(baseError.code.statusCode)(Json.toJson(baseError)),
-        response => Ok(Json.toJson(response))
-      )
+      withReusableSourceAndSize(request.body) {
+        (source, size) =>
```
drnaylor-oc, line 95: This is overdoing it a bit, because what's happening is that `Action.stream` is saving to a file, then this call is taking the file, reading it, then streaming it into another file. We should be able to fold this logic into `Action.stream` instead.

app/uk/gov/hmrc/transitmovements/utils/SmallMessageLimit.scala
akila-sankar: Continued with the service option

test/uk/gov/hmrc/transitmovements/controllers/MovementsControllerSpec.scala
```diff
@@ -37,6 +37,8 @@ import org.scalatest.matchers.must.Matchers
 import org.scalatestplus.scalacheck.ScalaCheckDrivenPropertyChecks
 import play.api.http.HeaderNames
 import play.api.http.MimeTypes
+import uk.gov.hmrc.objectstore.client.ObjectSummaryWithMd5
+import uk.gov.hmrc.objectstore.client.Path
```
drnaylor-oc, line 41: nitpick: make imports alphabetical please (except for scala/java imports, which go below in their own separate block)

test/uk/gov/hmrc/transitmovements/controllers/MovementsControllerSpec.scala
```diff
@@ -238,6 +255,64 @@ class MovementsControllerSpec
...
+        mockMessageFactory.createSmallMessage(
+          any[String].asInstanceOf[MessageId],
+          any[MessageType],
+          any[OffsetDateTime],
+          any[OffsetDateTime],
+          any[Option[MessageId]],
+          any[String].asInstanceOf[ObjectStoreResourceLocation],
+          any[MessageStatus]
+        )
+      ).thenReturn(message)
```
drnaylor-oc, line 299: We've been having a few issues with `any` matchers in some of our tests -- this morning we discovered that a typo had resulted in a broken system, partially because a matcher was set to `anyString()` when it could have been set to `eqTo("common-transit-convention-traders")`

Some of these should remain `any` because they are random by nature (message ID/movement ID in your case), but some of these possibly could be matched to an exact value, such as `MessageType.DeclarationData`

test/uk/gov/hmrc/transitmovements/services/ObjectStoreServiceSpec.scala
```diff
@@ -110,6 +119,64 @@ class ObjectStoreServiceSpec
...
+      val path: java.nio.file.Path      = file.toPath
+      val source: Source[ByteString, _] = Source.single(ByteString("this is test content")) //FileIO.fromPath(path)
+
+      when(
+        mockObjectStoreClient.putObject(
+          any[File],
+          any[Source[ByteString, _]],
+          any[RetentionPeriod],
+          any[Option[String]],
+          any[Option[Md5Hash]],
```
drnaylor-oc, line 138: We know the retention period (7 years), source (`source`) and MD5 hash (from `objectSummary`), so we should ensure they're being used here by using `eqTo`.

test/uk/gov/hmrc/transitmovements/services/SmallMessageLimitServiceSpec.scala
```diff
@@ -0,0 +1,45 @@
...
+package uk.gov.hmrc.transitmovements.services
+
+import org.scalatest.freespec.AnyFreeSpec
+import org.scalatest.matchers.must.Matchers.convertToAnyMustWrapper
+import org.scalatestplus.mockito.MockitoSugar
+import uk.gov.hmrc.transitmovements.config.AppConfig
+
+class SmallMessageLimitServiceSpec extends AnyFreeSpec with MockitoSugar {
+
+  private val config: AppConfig = mock[AppConfig]
```
drnaylor-oc, line 26: Define the limit, else it defaults to zero and I'd argue that's not really meaningful in the context of what you're trying to test.

drnaylor-oc (commented):
Nothing earth shattering, mostly test hygiene.

app/uk/gov/hmrc/transitmovements/controllers/MovementsController.scala
```diff
@@ -82,38 +88,97 @@ class MovementsController @Inject() (
...
+        if (smallMessageLimitService.checkContentSize(size)) createArrivalWithObjectStore(eori, request)
+        else {
+          for {
+            arrivalData <- movementsXmlParsingService.extractArrivalData(request.body).asPresentation
+            received   = OffsetDateTime.ofInstant(clock.instant, ZoneOffset.UTC)
+            movementId = MovementId(ShortUUID.next(clock, random))
+            message <- messageFactory
+              .create(MessageType.ArrivalNotification, arrivalData.generationDate, received, None, request.body, MessageStatus.Processing)
+              .asPresentation
+            movement = movementFactory.createArrival(movementId, eori, MovementType.Arrival, arrivalData, message, received, received)
```
liandra-giacomine, line 104: why are we creating the movement id here rather than in the factory?

app/uk/gov/hmrc/transitmovements/controllers/MovementsController.scala
```diff
@@ -82,38 +88,97 @@ class MovementsController @Inject() (
...
-          .asPresentation
-        movement = movementFactory.createDeparture(eori, MovementType.Departure, declarationData, message, received, received)
-        _ <- repo.insert(movement).asPresentation
-      } yield MovementResponse(movement._id, Some(movement.messages.head.id))).fold[Result](
+  private def createArrivalWithObjectStore(
+    eori: EORINumber,
+    request: Request[Source[ByteString, _]]
+  )(implicit hc: HeaderCarrier): EitherT[Future, PresentationError, MovementResponse] =
+    for {
+      arrivalData <- movementsXmlParsingService.extractArrivalData(request.body).asPresentation
```
liandra-giacomine, line 116: insert in the repo first, use the returned object to then pass the ids to object store service

app/uk/gov/hmrc/transitmovements/services/ObjectStoreService.scala
```diff
@@ -62,4 +76,31 @@ class ObjectStoreServiceImpl @Inject() (client: PlayObjectStoreClient) extends O
...
+    source: Source[ByteString, _]
+  )(implicit ec: ExecutionContext, hc: HeaderCarrier): EitherT[Future, ObjectStoreError, ObjectSummaryWithMd5] =
+    EitherT {
+      val formattedDateTime = dateTimeFormatter.format(OffsetDateTime.ofInstant(clock.instant, ZoneOffset.UTC))
+
+      (for {
+        response <- client.putObject(
+          path = Path.Directory(s"movements/${movementId.value}").file(s"${movementId.value}-${messageId.value}-$formattedDateTime.xml"),
+          content = source,
+          owner = "common-transit-conversion-traders"
```
sonisn1988, line 94: type should be ``` common-transit-convention-traders ```

drnaylor-oc: use the constant here: https://github.com/hmrc/transit-movements/blob/c9f8490c97a78e484cf01937dd811ecd189e32d3/app/uk/gov/hmrc/transitmovements/models/ObjectStoreURI.scala#L25

app/uk/gov/hmrc/transitmovements/controllers/MovementsController.scala
akila-sankar: We already had a discussion and decided to go with this approach.

liandra-giacomine: Cool, just for my understanding  - what was the logic behind the decision? Because if it doesn't save to object-store then we won't be saving it in the DB?

drnaylor-oc: To be fair, I did say "if that's the best way" and I'm not as close to the code as I once was -- I don't want to be totally prescriptive in how we do this.

In the context of small messages -- it is a bit chicken and egg. If we put it in the DB first then fail the call to object store, then we just need to ensure the status gets set to failed in the DB. If we put it in object store first, we need to pull id generation out.

From the point of view of retention, if we put it in OS then it'll hang around for 7 years, if we start with the DB that's 31 days. That might be a factor in decisions.

Finally, ideally, the controller wouldn't generate the IDs but I understand why in this case. I mentioned that reservation at the time we discussed it.

So, either approach is fine, but we do need to agree on it. Note that I do need to do some work in transit-movements around object store for platops anyway, so worst comes to worst we can tweak it later.

I do want to do it right, but remember we're agile too!

liandra-giacomine: If we put it in object-store first, then we would need to not fail the for comprehension if the call to object-store fails. Right now it won't create a movement in the DB if the call fails so just change `objectSummary <- objectStoreService.addMessage(movementId, messageId, request.body).asPresentation` to `objectSummary = objectStoreService.addMessage(movementId, messageId, request.body).asPresentation`

drnaylor-oc: For the small message route, that's okay though, because this will be from a sync call -- so if there is no DB entry, it doesn't matter because the trader should already know.

akila-sankar: Agree

derekspalmer (approved):

PR #60, opened 2023-03-17
CTCP-2757: Fix typo from a previous commit, update tests, update token

drnaylor-oc:


sonisn1988 (approved):

conf/application.conf
```diff
@@ -129,4 +129,4 @@ microservice {
 }
 
 object-store.default-retention-period = 1-month
-internal-auth.token = "1234"
+internal-auth.token = "transit-movements-token"
```
sonisn1988, line 132: I think this token is going to be same as ``` common-transit-convention-traders-token ``` because resource location start with  in transit-movements ``` common-transit-convention-traders-token ```

drnaylor-oc: The token is per service, each token will (currently) get this grant. So they should NOT be the same.

PR #61, opened 2023-03-21
CTCP-2379, CTCP-2751: Fix message ID being duplicated and causing mismatches in database, fix use of ObjectStoreResourceLocation

drnaylor-oc:
Also tightens tests to ensure the right things go in and out.

Will write a tech debt ticket to rebuild the test class as it's using a lot of global constants when it should all be test local.


app/uk/gov/hmrc/transitmovements/controllers/MovementsController.scala
```diff
@@ -125,7 +122,7 @@ class MovementsController @Inject() (
           arrivalData.generationDate,
           received,
           None,
-          ObjectStoreResourceLocation(objectSummary.location.asUri),
+          ObjectStoreURI(objectSummary.location.asUri),
```
derekspalmer, line 125: This seems to flip-flop between the 2?  I think we have ObjectStoreResourceLocation in common-transit-convention-traders-

derekspalmer (approved):
Apart from test failure, looks fine.

app/uk/gov/hmrc/transitmovements/controllers/MovementsController.scala
drnaylor-oc: It should be both:

* `ObjectStoreURI` contains the entire URI to object-store
* `ObjectStoreResourceLocation` is basically the same, but with the owner.

When this was introduced, it's been used for both the URI and the owner-less URI, and it was causing issues with storing and retrieving. The whole idea of using the `AnyVal` wrappers is to ensure the right string goes to the right place, but unfortunately, due to the fact this was used in two different ways, I split them like this.

Because the object-store code has to be rationalised into `transit-movements` anyway, that problem should go away elsewhere.

PR #62, opened 2023-03-24
[CTCP-2451] added support to accept optional messageType for PATCH endpoint updateMessage

prashanthmamidi:


app/uk/gov/hmrc/transitmovements/controllers/MovementsController.scala
```diff
@@ -391,4 +394,30 @@ class MovementsController @Inject() (
...
+        case Some(messageType) =>
+          movementType match {
+            case MovementType.Arrival =>
+              MessageType.checkArrivalMessageType(messageType.code) match {
+                case None    => Future.successful(Left(InvalidMessageType(s"Invalid messageType value: $messageType")))
+                case Some(_) => Future.successful(Right())
+              }
+            case MovementType.Departure =>
+              MessageType.checkDepartureMessageType(messageType.code) match {
+                case None =>
```
liandra-giacomine, line 412: The message type should be optional, if not present then we just keep going rather than return an error.

This can also be condensed to:

`(messageType, movementType) match {
case (Some(messageType), Arrival) => //check and return appropriate result
case (Some(messageType), Departure) => //check and return appropriate result
case _ => Future.successful(Right())
}`

prashanthmamidi: Yes, if messageType is None, then we return  `case None => Future.successful(Right())`

The code you have mentioned, we return error when we try to update Arrival movement with departure messageType and also we return error when we try to update Departure movement with arrival messageType

liandra-giacomine: I think it's the same logic just condensed:

`    (messageType, movementType) match {
         case (Some(messageType), MovementType.Arrival) =>
               MessageType.checkArrivalMessageType(messageType.code) match {
                 case None    => Future.successful(Left(InvalidMessageType(s"Invalid messageType value: $messageType")))
                 case Some(_) => Future.successful(Right())
               }
            case (Some(messageType), MovementType.Departure) =>
               MessageType.checkDepartureMessageType(messageType.code) match {
                 case None =>
                   Future.successful(Left(InvalidMessageType(s"Invalid messageType value: $messageType")))
                 case Some(_) =>
                   Future.successful(Right())
               }
           }

         case None => Future.successful(Right())
     }`

liandra-giacomine (approved):

PR #67, opened 2023-04-03
CTCP 2571: Update add message route to support sending messages to Object Store

drnaylor-oc:
Part 2 of the object-store rationalisation. Another part is coming later.

# What is this PR all about?

In the previous PR, I added support for storing messages from all routes, except for two:

* Attaching a new message from the router (that process required a file to be added to object-store first
* Attaching a message body to an empty movement via a stream

This PR will do the first of these

# What does this PR do?

* Sending a stream to `/traders/movements/:movementId/messages` can now be stored in object store automatically, based on file size
* Selection of where a message body goes is now decided by the `MessageService` (renamed from `MessageFactory`)
* Message size has been added to the mongo database as this will be useful later
* Some de-duplication has occurred as a result

# What's next?

* Adding a POST message body endpoint for attaching a body to support small messages becoming large messages


app/uk/gov/hmrc/transitmovements/controllers/MovementsController.scala
```diff
@@ -196,10 +142,10 @@ class MovementsController @Inject() (
       )
   }
 
-  def updateMovement(movementId: MovementId, triggerId: Option[MessageId] = None) =
+  def updateMovement(movementId: MovementId, triggerId: Option[MessageId] = None): Action[Source[ByteString, _]] =
     contentTypeRoute {
-      case Some(_) => updateMovementSmallMessage(movementId, triggerId)
-      case None    => updateMovementLargeMessage(movementId, triggerId)
+      case Some(_) => updateMovementWithStream(movementId, triggerId)
```
derekspalmer, line 147: Yep, more descriptive  :+1:

derekspalmer (approved):

PR #68, opened 2023-04-03
CTCP-2571, CTCP-2452: Add create body endpoint

drnaylor-oc:
Part 3 of the object-store rationalisation, to support CTCP-2452

# What is this PR all about?

Attaching a message body to an empty message via a stream

# What does this PR do?

* Sending a stream to `POST /traders/:EORI/movements/:movementType/:movementId/messages/:messageId/body` will add a message body to an empty movement.
    * 201 on creation
    * 404 if the message ID or movement ID doesn't exist
    * 409 (conflict) if the message body already exists (this is why we're using POST, it's not idempotent)
    * 400 if the XML is bad
    * 500 if something goes wrong
* Message status on `MessageResponse` is now optional
* Retains `UpdateMessageMetadata` solely as a presentation layer object, `UpdateMessageData` is now the data transfer object to Mongo

# What's next?

This should be it for `transit-movements` for now, apart from supporting receiving upscan URLs instead of object-store, but that's not a blocker right this moment. Next, once this has settled, we'll go around other repos and point them at better places here instead of object store, then change the owner of everything from `common-transit-convention-traders` to `transit-movements`.

### DIFF
--- a/app/uk/gov/hmrc/transitmovements/controllers/MovementsController.scala
+++ b/app/uk/gov/hmrc/transitmovements/controllers/MovementsController.scala
@@ -28,7 +28,6 @@ import play.api.libs.json.Reads
 import play.api.mvc.Action
 import play.api.mvc.AnyContent
 import play.api.mvc.ControllerComponents
-import play.api.mvc.Request
 import play.api.mvc.Result
 import uk.gov.hmrc.http.HeaderCarrier
 import uk.gov.hmrc.play.bootstrap.backend.controller.BackendController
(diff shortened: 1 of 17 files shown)


drnaylor-oc:
Note that this requires #67 to merge first

app/uk/gov/hmrc/transitmovements/controllers/MovementsController.scala
```diff
@@ -313,7 +317,7 @@ class MovementsController @Inject() (
           source           <- objectStoreService.getObjectStoreFile(resourceLocation).asPresentation
           extractedData    <- movementsXmlParsingService.extractData(movementType, source).asPresentation
           _ <- repo
-            .updateMovement(movementId, Some(extractedData.movementEoriNumber), extractedData.movementReferenceNumber, extractedData.generationDate)
+            .updateMovement(movementId, Some(extractedData.movementEoriNumber), extractedData.movementReferenceNumber, receieved)
```
derekspalmer, line 320: typo receieved

derekspalmer (approved):
LGTM

derekspalmer (approved):

PR #71, opened 2023-04-24
Modified identification number to be optional during xml extraction

liandra-giacomine:


app/uk/gov/hmrc/transitmovements/services/XmlParsers.scala
```diff
@@ -32,13 +32,18 @@ import java.time.format.DateTimeParseException
...
 object XmlParsers extends XmlParsingServiceHelpers {
 
-  def movementEORINumberExtractor(rootNode: String, eori: String): Flow[ParseEvent, ParseResult[EORINumber], NotUsed] =
+  def movementEORINumberExtractor(rootNode: String, eori: String): Flow[ParseEvent, ParseResult[Option[EORINumber]], NotUsed] =
     XmlParsing
       .subtree(rootNode :: eori :: "identificationNumber" :: Nil)
       .collect {
         case element if element.getTextContent.nonEmpty => EORINumber(element.getTextContent)
       }
       .single("identificationNumber")
```
drnaylor-oc, line 41: `single` is an extension method. Could we just create another method entitled `singleOption` and use that, rather than having to match after it?

liandra-giacomine: Yes that'd be better! will do

drnaylor-oc (approved):
Passed tests, just waiting for a good time to merge.

PR #72, opened 2023-04-27
CTCP-2765 Get movement message mongo respository changes

akila-sankar:


app/uk/gov/hmrc/transitmovements/repositories/MovementsRepository.scala
```diff
@@ -185,9 +185,13 @@ class MovementsRepositoryImpl @Inject() (
 
     mongoRetry(Try(collection.aggregate[MovementWithoutMessages](aggregates)) match {
       case Success(obs) =>
-        obs.headOption().map {
-          opt => Right(opt)
-        }
+        obs
+          .head()
```
drnaylor-oc, line 189: Better to use `headOption` here still:

```scala
obs.headOption().map {
    case Some(x) => Right(x)
    case None    => Left(DocumentNotFound(s"No movement found with the given id: ${movementId.value}"))
}
```

I'm surprised `obs.head()` returns `null`, but we should still use `headOption()` to make it clear we might not get something back.

app/uk/gov/hmrc/transitmovements/repositories/MovementsRepository.scala
```diff
@@ -250,8 +254,10 @@ class MovementsRepositoryImpl @Inject() (
 
     mongoRetry(Try(collection.aggregate[MessageResponse](aggregates)) match {
       case Success(obs) =>
-        obs.headOption().map {
-          opt => Right(opt)
+        obs.head().map {
+          opt =>
+            if (opt != null) Right(opt)
+            else Left(DocumentNotFound(s"Message ID ${messageId.value} for movement ID ${movementId.value} was not found"))
```
drnaylor-oc, line 260: As above.

akila-sankar: Yes updated

drnaylor-oc:
Have you tried running the regression tests against this? All being well, a few should fail because of the fact we're changing a 404 to a 200 (correctly so), in which case we'll need to ensure that we have that ready to go too before this merges.

akila-sankar:
> Have you tried running the regression tests against this? All being well, a few should fail because of the fact we're changing a 404 to a 200 (correctly so), in which case we'll need to ensure that we have that ready to go too before this merges.

Yes @drnaylor-oc regression tests are not impacted, because earlier it was handled in controller to return 404 & now it is moved to repository layer.

drnaylor-oc:
Oh, yes, I'm thinking of a different ticket -- yup, you're right! Ignore me!

akila-sankar:
> Oh, yes, I'm thinking of a different ticket -- yup, you're right! Ignore me!

Sure, no problem

PR #73, opened 2023-05-15
CTCP-2949 Restrict LRN with message sender combination for IE015

akila-sankar:


app/uk/gov/hmrc/transitmovements/repositories/MovementsRepository.scala
```diff
@@ -143,7 +147,8 @@ class MovementsRepositoryImpl @Inject() (
         Codecs.playFormatCodec(MongoFormats.movementIdFormat),
         Codecs.playFormatCodec(MongoFormats.mrnFormat),
         Codecs.playFormatCodec(MongoFormats.offsetDateTimeFormat),
-        Codecs.playFormatCodec(MongoFormats.eoriNumberFormat)
+        Codecs.playFormatCodec(MongoFormats.eoriNumberFormat),
+        Codecs.playFormatCodec(MongoFormats.lrnFormat)
       )
```
drnaylor-oc, line 152: Can we add the LRN as an index as we're going to scan it all the time? If possible, can we mark it as a unique index too? If so, would it be possible to avoid the separate check we do (more thinking if we're rejected due to a duplicate, can we detect that in code and send the right error message?)

drnaylor-oc (commented):
One thing in the code I spotted -- plus we should put an index on the database for the LRN given we'll be scanning that all the time.

app/uk/gov/hmrc/transitmovements/repositories/MovementsRepository.scala
```diff
@@ -421,4 +442,31 @@ class MovementsRepositoryImpl @Inject() (
...
+      Aggregates.filter(selector),
+      Aggregates.sort(descending("updated")),
+      Aggregates.project(projection)
+    )
+
+    mongoRetry(Try(collection.aggregate[MovementWithoutMessages](aggregates)) match {
+      case Success(obs) =>
+        obs.headOption().map {
+          case None =>
+            Right()
```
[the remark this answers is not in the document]
akila-sankar: Sure, will update it

app/uk/gov/hmrc/transitmovements/repositories/MovementsRepository.scala
drnaylor-oc: Okay, in that case just add the index but don't make it unique because we still need fast read access. We'll keep the selector you've written in to do the duplicate check.

akila-sankar: Yes agree, will add the index

app/uk/gov/hmrc/transitmovements/controllers/MessageBodyController.scala
```diff
@@ -94,7 +96,16 @@ class MessageBodyController @Inject() (
...
           extractedData <- movementsXmlParsingService.extractData(messageType, request.body).asPresentation
-          bodyStorage   <- messageService.storeIfLarge(movementId, messageId, size, request.body).asPresentation
+          _ <- messageType match {
+            case MessageType.DeclarationData =>
+              repo
+                .restrictDuplicateLRN(
+                  extractedData.flatMap(_.localReferenceNumber).get
+                )
+                .asPresentation
+            case _ => EitherT.rightT[Future, MongoError]((): Unit).asPresentation
```
drnaylor-oc, line 106: Not convinced we should do this right now, but I did want to get my thoughts down to possibly sort this later as part of the object-store rationalisation stuff.

---

At some point I guess I'm splitting hairs, but the LRN is only extracted for declaration data anyway, and for neatness it would be better to keep pattern matching out of this comprehension, which is a series of steps that is easy to read.

Perhaps create a method like so (implicits may need sorting out too):

```scala
def checkForDuplicateLRN(movement: Option[Movement]): EitherT[Future, MongoError, Unit] = 
    message match {
        case Some(Movement(_, _, _, _, _, Some(lrn), _, _, _)) => repo.restrictDuplicateLRN(lrn)
        case None => EitherT.rightT[Future, MongoError]((): Unit)
    }
``` 

then just use this method in the for comprehension.

We have to do this in the `MovementsController` too and this should really be part of a service, hence why I'm not going to insist on this right now

drnaylor-oc (approved):
I'm relatively happy with this. I think there is a possible improvement here, but it's more stylistic, and it might get moved when I complete the object-store rationalisation.

So you can do this, but equally feel free to ignore me for now.

Merge once we have all the tests ready.

drnaylor-oc (dismissed):
Reapproving due to branch change. Original comments still apply.

app/uk/gov/hmrc/transitmovements/controllers/MessageBodyController.scala
akila-sankar: Thank you @drnaylor-oc for the suggestion, as it might get moved after the object-store rationalisation I better leave it here.

drnaylor-oc (approved):

PR #75, opened 2023-05-22
[CTCP-3079] added support to filter by movementReferenceNumber

prashanthmamidi:


app/uk/gov/hmrc/transitmovements/repositories/MovementsRepository.scala
```diff
@@ -295,16 +297,31 @@ class MovementsRepositoryImpl @Inject() (
...
+                  Right(
+                    response
+                      .filter(
+                        movement => partialMrnMatch(movementReferenceNumber, movement)
+                      )
+                      .toVector
+                  )
+                case _ => Right(response.toVector)
+              }
+          }
```
drnaylor-oc, line 313: This can be done in the database using the regex. You'll want `/\Q{string}\E/`, possibly with the case sensitive identifier.

See https://www.mongodb.com/docs/manual/reference/operator/query/regex/

If a trader has thousands of movements, this is going to be expensive. The database is built for it -- we should use that.

prashanthmamidi: yes, it's better to use mongo filter - tried this earlier - it doesn't work 
https://github.com/hmrc/transit-movements/pull/75/commits/900dfc8e0b98be583f9ef3c8eeb9e39106a590c4#diff-c2d588a06817157d7dd8e752554d6cd3454e2524f60893a9acc904fdc441cccbR335

Let me try as suggested `s"/\\Q${movementReferenceNumber.value}\\E/"` and see how it goes

prashanthmamidi: thanks - yes this filter `s"\\Q${movementReferenceNumber.value}\\E"` works well

derekspalmer (approved):

derekspalmer (approved):

derekspalmer (approved):

PR #79, opened 2023-05-31
Incorrect status code returned when filtering returns no messages/mov…

sonisn1988:
…ements

### DIFF
--- a/test/uk/gov/hmrc/transitmovements/controllers/MovementsControllerSpec.scala
+++ b/test/uk/gov/hmrc/transitmovements/controllers/MovementsControllerSpec.scala
@@ -27,7 +27,6 @@ import org.mockito.Mockito.times
 import org.mockito.Mockito.verify
 import org.mockito.MockitoSugar.reset
 import org.mockito.MockitoSugar.when
-import org.mockito.MockitoSugar.when
 import org.scalacheck.Arbitrary.arbitrary
 import org.scalacheck.Gen
 import org.scalatest.BeforeAndAfterEach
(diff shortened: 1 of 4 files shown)


test/uk/gov/hmrc/transitmovements/controllers/MovementsControllerSpec.scala
```diff
@@ -862,14 +863,15 @@ class MovementsControllerSpec
...
           (updatedSince, movementEORI) =>
             when(mockRepository.getMovements(EORINumber(any()), eqTo(movementType), eqTo(updatedSince), eqTo(movementEORI), eqTo(None)))
               .thenReturn(EitherT.rightT(Vector.empty[MovementWithoutMessages]))
 
             val result = controller.getMovementsForEori(eoriNumber, movementType, updatedSince, movementEORI)(request)
 
-            status(result) mustBe NOT_FOUND
+            status(result) mustBe OK
+            contentAsJson(result) mustBe Json.toJson(Vector.empty[MovementWithoutMessages])
         }
```
drnaylor-oc, line 875: Can we make sure we have a test that test that we get `NOT_FOUND` if a **movement** is not found (rather than a valid movement with no messages?)

Can we check that the Mongo repository layer does this? Is there a test we can write for this, or do they already exist?

drnaylor-oc (approved):
As with the frontend, await LRN and regression tests to be sorted.

drnaylor-oc (approved):
As with https://github.com/hmrc/common-transit-convention-traders/pull/432#pullrequestreview-1465099883

PR #80, opened 2023-06-05
[CTCP-3078] added support for lrn filters

prashanthmamidi:


sonisn1988 (approved):
LGTM

sonisn1988 (dismissed):

drnaylor-oc:
I had a look at the test you commented out and went ahead and make it more explicit here https://github.com/hmrc/transit-movements/pull/84. However, when I ran the original test itself, it ran in its own database and collection -- so while I thought I understood why you did it, it all worked for me without wiping out other databases.

Regardless, we shouldn't comment out tests, or part of tests -- if there is a problem with one, we should look into it.

I'd like to get my PR in, and then you use that as your base, rather than have the test semi commented out.

prashanthmamidi:
> I had a look at the test you commented out and went ahead and make it more explicit here #84. However, when I ran the original test itself, it ran in its own database and collection -- so while I thought I understood why you did it, it all worked for me without wiping out other databases.
> 
> Regardless, we shouldn't comment out tests, or part of tests -- if there is a problem with one, we should look into it.
> 
> I'd like to get my PR in, and then you use that as your base, rather than have the test semi commented out.

thanks - was fixing quite a few merge issues and temporarily commented this out as this failing in my local to see first my merge went on well. Will incorporate your changes here for this test.

drnaylor-oc:
Given you've copied it in here, let's just approve and merge this once we get a chance, no need for my PR now!

drnaylor-oc (approved):

PR #82, opened 2023-06-12
Test-only endpoint to drop movements collection.

jackheslop96:


drnaylor-oc:
Just to reiterate what I said for everyone's benefit, I don't have a specific issue with this, but if we can do this directly from Gatling scripts without needing this, that would be better.

Else, almost want to say "add a test!", if only so we keep the coverage up!

drnaylor-oc (approved):
Given End to End testing is ongoing, we're being careful about when we merge. I'll try to get this merged around lunch time.

As for configuration, let's put this in `app-config-staging` only. I'd rather not have to forget about not putting it into production and external test!

jackheslop96:
> Given End to End testing is ongoing, we're being careful about when we merge. I'll try to get this merged around lunch time.
> 
> As for configuration, let's put this in `app-config-staging` only. I'd rather not have to forget about not putting it into production and external test!

Thanks @drnaylor-oc - here's the config branch if you could give it the once over? Thanks - https://github.com/hmrc/app-config-staging/pull/13057

PR #91, opened 2023-07-14
CTCP-3480 LRN + message sender combo for duplicate check

akila-sankar:


app/uk/gov/hmrc/transitmovements/repositories/MovementsRepository.scala
```diff
@@ -525,8 +534,8 @@ class MovementsRepositoryImpl @Inject() (
         Future.successful(Left(UnexpectedError(Some(ex))))
     })
 
-  def restrictDuplicateLRN(localReferenceNumber: LocalReferenceNumber): EitherT[Future, MongoError, Unit] = {
-    val selector = mEq("localReferenceNumber", localReferenceNumber.value)
+  def restrictDuplicateLRN(localReferenceNumber: LocalReferenceNumber, messageSender: MessageSender): EitherT[Future, MongoError, Unit] = {
+    val selector = mAnd(mEq("localReferenceNumber", localReferenceNumber.value), mEq("messageSender", messageSender.value))
```
drnaylor-oc, line 538: This query will need to be expanded on slightly -- from the ticket:

> ... OR if the LRN has been used in a departure with no message sender, so that we still catch all duplicates

So, this query needs to return if there are any entries if:

* the LRN and messageSender exist on another document (which is this document), OR
* the LRN exists on a document WITHOUT a messageSender (so `mAnd(mEq("localReferenceNumber"), mNot(mExists("messageSender")))`)

Note that `mNot` and `mExists` would need to be aliased from `Filters.not` and `Filters.exists`.

We need this as the database doesn't currently hold message senders, so we don't want to let potential duplicate LRNs through.

akila-sankar: Yes correct, updated it please check

drnaylor-oc:
Is there an integration test for testing for a duplicate LRN if said LRN has been used but without a message sender (or with)? Similarly, is there one for not flagging a duplicate if the LRN is duplicated but the message sender is not? I don't see any additional having been written and this is kind of new functionality.

akila-sankar:
Added now for the scenario of duplicate LRN without messageSender & for all the possible scenarios with messageSender I had already updated IT tests.

sonisn1988 (approved):
LGTM. We can merge PR once regression is ready and approve for this.

VijayKumarTech (approved):
All the tests on /ctc-traders-api-tests/pull/155 are passing against this branch.
LGTM.

PR #92, opened 2023-08-22
exclude messages field for movements

sonisn1988:


drnaylor-oc (approved):
I'm okay with this as long as the project tests and regression tests are still passing with it. Given the use of facet (which we now know more about!) this is a stop-gap until we fix it properly.

PR #93, opened 2023-08-24
CTCP-3705: Add configuration for disabling LRN check

drnaylor-oc:
For now, rather than remove the check, we just need to be able to not do it on a per-environment basis -- so we have a config option to toggle when the time is right.

We will want to delete this function later, but when everything is working up to ET.


drnaylor-oc:
Just run all regression tests which passed, so merging.

PR #94, opened 2023-08-25
Remove the use of “facet” when getting a list of movements/messages f…

sonisn1988:
…rom Mongo

### DIFF
--- a/app/uk/gov/hmrc/transitmovements/repositories/MovementsRepository.scala
+++ b/app/uk/gov/hmrc/transitmovements/repositories/MovementsRepository.scala
@@ -32,31 +32,14 @@ import com.mongodb.client.model.UpdateOptions
 import com.mongodb.client.model.Updates.{combine => mCombine}
 import com.mongodb.client.model.Updates.{push => mPush}
 import com.mongodb.client.model.Updates.{set => mSet}
-
 import org.bson.conversions.Bson
-import org.mongodb.scala.bson.Document
 import org.mongodb.scala.model.Sorts.descending
 import org.mongodb.scala.model._
 import play.api.Logging
-import play.api.libs.json.Json
 import uk.gov.hmrc.mongo.MongoComponent
 import uk.gov.hmrc.mongo.play.json._
 import uk.gov.hmrc.transitmovements.config.AppConfig
-import uk.gov.hmrc.transitmovements.models.EORINumber
-import uk.gov.hmrc.transitmovements.models.ItemCount
-import uk.gov.hmrc.transitmovements.models.LocalReferenceNumber
-import uk.gov.hmrc.transitmovements.models.Message
-import uk.gov.hmrc.transitmovements.models.MessageId
-import uk.gov.hmrc.transitmovements.models.MessageSender
-import uk.gov.hmrc.transitmovements.models.Movement
-import uk.gov.hmrc.transitmovements.models.MovementId
-import uk.gov.hmrc.transitmovements.models.MovementReferenceNumber
-import uk.gov.hmrc.transitmovements.models.MovementType
-import uk.gov.hmrc.transitmovements.models.MovementWithoutMessages
-import uk.gov.hmrc.transitmovements.models.PageNumber
-import uk.gov.hmrc.transitmovements.models.PaginationMessageSummary
-import uk.gov.hmrc.transitmovements.models.PaginationMovementSummary
-import uk.gov.hmrc.transitmovements.models.UpdateMessageData
+import uk.gov.hmrc.transitmovements.models._
 import uk.gov.hmrc.transitmovements.models.formats.CommonFormats
 import uk.gov.hmrc.transitmovements.models.formats.MongoFormats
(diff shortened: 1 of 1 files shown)


app/uk/gov/hmrc/transitmovements/repositories/MovementsRepository.scala
```diff
@@ -340,51 +326,52 @@ class MovementsRepositoryImpl @Inject() (
...
       case Failure(NonFatal(ex)) =>
         Future.successful(Left(UnexpectedError(Some(ex))))
     })
 
-  }
+  private def totalCountMovements(totalDocument: Seq[Bson]): EitherT[Future, MongoError, Long] =
+    mongoRetry(Try(collection.aggregate[MovementWithoutMessages](totalDocument)) match {
+      case Success(obs) =>
+        obs.toFuture().map {
+          response => Right(response.size.toLong)
```
drnaylor-oc, line 360: Rather than pulling all the documents into memory and measuring the size of the list, use the `$count` aggregation instead. 

You could put this in this method, so that you pass the filter you want to use in, and this method automatically puts the `count` into the pipeline if you so wished.

app/uk/gov/hmrc/transitmovements/repositories/MovementsRepository.scala
```diff
@@ -340,51 +326,52 @@ class MovementsRepositoryImpl @Inject() (
...
-                "$arrayElemAt" -> Json.arr("$totalCount", 0)
-              )
-            ),
-            "in" -> "$$countValue.count"
-          )
-        ),
-        0
-      )
-    )
+  private def totalCountMessages(totalDocument: Seq[Bson]): EitherT[Future, MongoError, Long] =
```
drnaylor-oc, line 366: This is just a duplicate of the above function except for the type that is returned, but by using the `$count` aggregation, that distinction will disappear too.

I suggest removing this method and renaming the other one to something like "countItems".

app/uk/gov/hmrc/transitmovements/repositories/MovementsRepository.scala
```diff
@@ -340,51 +326,52 @@ class MovementsRepositoryImpl @Inject() (
...
+    for {
+      perPageMovements <- filterPerPageMovements(aggregates)
+      totalCount       <- totalCountMovements(filterAggregates)
+    } yield PaginationMovementSummary(TotalCount(totalCount), perPageMovements)
+
+  }
 
-    mongoRetry(Try(collection.aggregate[PaginationMovementSummary](aggregates)) match {
+  private def filterPerPageMovements(aggregates: Seq[Bson]): EitherT[Future, MongoError, Vector[MovementWithoutMessages]] =
+    mongoRetry(Try(collection.aggregate[MovementWithoutMessages](aggregates)) match {
```
drnaylor-oc, line 344: This method duplicates `filterPerPageMessages` apart from the return type -- rather than duplicate code, use a generic:

```
private def filterPerPage[R](aggregates: Seq[Bson]): EitherT[Future, MongoError, Vector[R]]
```

It might need an implicit defining too.

drnaylor-oc (commented):
Biggest issue I have is that it still pulls back everything we have in the database that fits the filter, we should avoid that and let the database optimise the counting too.

app/uk/gov/hmrc/transitmovements/repositories/MovementsRepository.scala
```diff
@@ -334,41 +336,24 @@ class MovementsRepositoryImpl @Inject() (
...
-  private def totalCountMessages(totalDocument: Seq[Bson]): EitherT[Future, MongoError, Long] =
-    mongoRetry(Try(collection.aggregate[MessageResponse](totalDocument)) match {
-      case Success(obs) =>
-        obs.toFuture().map {
-          response => Right(response.size.toLong)
-        }
+          .map(
+            _.headOption
+              .map(
+                bsonDocument => Right(bsonDocument.get("count").asInt32().longValue())
```
drnaylor-oc, line 353: `asInt32().longValue()` negates the point of using a `Long`. Can the `asInt32()` be something like `asInt64()` instead, as a `Long` is 64 bits?

(I don't know if that exists, but converting the number to a 32 bit number then immediately to a 64 bit number makes no sense)

drnaylor-oc: Use `asNumber`.

I looked into how $sum works (which is basically what $count uses) and it's basically that grouping uses the smallest data type it can, `Int32`. https://www.mongodb.com/docs/v7.0/reference/operator/aggregation/sum/#result-data-type suggests that it can be upgraded to `Int64` if necessary.

Now, I admit, it could all be moot as I don't expect a trader to have over 2.2 million movements active at once unless the TTL is massively increased, but as we're already using long we should at least ensure that it has purpose. `asNumber` will create a `BsonNumber` that could be either an `Int32`, `Int64` or `Double`, so that should work as expected and not lose this potential edge case.

drnaylor-oc (approved):

PR #102, opened 2024-05-16
CTCP-5297: bump up ttl to 60 days

derekspalmer:


conf/application.conf
```diff
@@ -74,7 +74,8 @@ metrics {
 mongodb {
   uri = "mongodb://localhost:27017/transit-movements"
   retryAttempts = 0
-  timeToLiveInSeconds = 2592000 # 30 days
+  timeToLiveInSeconds = 5184000 # 60 days
+  replaceIndexes = false
```
sonisn1988, line 78: is this required?

app/uk/gov/hmrc/transitmovements/repositories/MovementsRepository.scala
```diff
@@ -165,7 +165,8 @@ class MovementsRepositoryImpl @Inject() (
         Codecs.playFormatCodec(mongoFormats.paginationMovementSummaryFormat),
         Codecs.playFormatCodec(mongoFormats.paginationMessageSummaryFormat),
         Codecs.playFormatCodec(mongoFormats.sensitiveStringFormat)
-      )
+      ),
+      replaceIndexes = appConfig.replaceIndexes
```
sonisn1988, line 169: is this required?

derekspalmer: Yes, If we have an existing collection then for the indexes to take effect we need to enforce it.
Jack said we need to trigger this and also there's a slack thread in community-mongo also indicating that we need to do this.

conf/application.conf
sonisn1988: I did some research and found that is required please make it true  for local as well otherwise we will end up with exception.

sonisn1988 (approved):

PR #110, opened 2024-08-07
CTCP-5772: Remove routing / versioned packages

DamienButterworth:
This can be removed as it has been stated that version 2.1 is forward compatible with version 2.0 therefore no specific logic needs to be done in out DB to handle the two different versions


app/uk/gov/hmrc/transitmovements/services/SourceManagementService.scala
```diff
@@ -0,0 +1,76 @@
...
+import scala.concurrent.Future
+
+@ImplementedBy(classOf[SourceManagementServiceImpl])
+trait SourceManagementService {
+  def replicateRequestSource(request: Request[Source[ByteString, _]], count: Int): EitherT[Future, PresentationError, List[Source[ByteString, _]]]
+  def replicateSource(source: Source[ByteString, _], count: Int): EitherT[Future, PresentationError, List[Source[ByteString, _]]]
+  def calculateSize(source: Source[ByteString, _]): EitherT[Future, PresentationError, Long]
+}
+
+class SourceManagementServiceImpl @Inject() (implicit materializer: Materializer, executionContext: ExecutionContext) extends SourceManagementService {
```
sonisn1988, line 39: Please add the unit test class for this file.

sonisn1988 (approved):

PR #112, opened 2024-09-02
CTCP-5832: httpv2 

joecefnhedog:


conf/application.conf
```diff
@@ -22,6 +22,8 @@ appName = transit-movements
 # Default http client
 play.modules.enabled += "uk.gov.hmrc.play.bootstrap.HttpClientModule"
```
DamienButterworth, line 23: If the previous module is not longer being used it can be removed

PR #114, opened 2024-09-25
CTCP-5914: Dual header handle transitional

DamienButterworth:


app/uk/gov/hmrc/transitmovements/models/MovementWithoutMessages.scala
```diff
@@ -42,6 +43,7 @@ object MovementWithoutMessages {
       movement.movementReferenceNumber,
       movement.localReferenceNumber,
       movement.created,
-      movement.updated
+      movement.updated,
+      movement.isTransitional
```
sonisn1988, line 47: I don't think so it is required?

app/uk/gov/hmrc/transitmovements/models/mongo/read/MongoMovementSummary.scala
```diff
@@ -58,7 +60,8 @@ case class MongoMovementSummary(
       movementReferenceNumber,
       localReferenceNumber,
       created,
-      updated
+      updated,
+      isTransitional.getOrElse(true)
```
sonisn1988, line 64: I don't think so it is required? As this is for get call from common-transit-convention-traders?

app/uk/gov/hmrc/transitmovements/repositories/MovementsRepository.scala
```diff
@@ -216,8 +217,15 @@ class MovementsRepositoryImpl @Inject() (
         obs
           .headOption()
           .map {
-            case Some(opt) => Right(opt)
-            case None      => Left(DocumentNotFound(s"No movement found with the given id: ${movementId.value}"))
+            case Some(opt) =>
```
sonisn1988, line 220: I don't think so it is required? As this is for get call from common-transit-convention-traders?

app/uk/gov/hmrc/transitmovements/models/mongo/read/MongoMovementEori.scala
```diff
@@ -29,21 +29,24 @@ object MongoMovementEori {
...
+  isTransitional: Option[Boolean]
 ) {
 
   @transient lazy val movementWithEori: MovementWithEori =
     MovementWithEori(
       _id,
       enrollmentEORINumber,
-      clientId
+      clientId,
+      isTransitional.getOrElse(true)
```
sonisn1988, line 49: I assume you have tested this when we don't have this field in database and it's returning ```None```.  Not sure if it can return ```null``` and then when we use ```isTransitional.getOrElse``` will ends up with ```NullPointerException```

sonisn1988 (approved):

PR #141, opened 2026-03-06
CTCP6-slow-query-fix: optimize slow query

tudorsonycx:
- add compound indexes to cover potential scenarios (still needs more)

- eliminate redundant double querying in getMovements and getMessages methods

- ~remove redundant background index build option since it's set to true by default since MongoDB 4.2+~

### DIFF
--- a/app/uk/gov/hmrc/transitmovements/repositories/MovementsRepository.scala
+++ b/app/uk/gov/hmrc/transitmovements/repositories/MovementsRepository.scala
@@ -161,17 +161,28 @@ class MovementsRepositoryImpl @Inject() (
           Indexes.compoundIndex(
             Indexes.ascending("enrollmentEORINumber"),
             Indexes.ascending("movementType"),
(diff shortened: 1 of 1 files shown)


app/uk/gov/hmrc/transitmovements/repositories/MovementsRepository.scala
```diff
@@ -154,24 +154,32 @@ class MovementsRepositoryImpl @Inject() (
...
+            Indexes.ascending("movementType"),
+            Indexes.ascending("localReferenceNumber"),
+            Indexes.ascending("updated")
+          )
         ),
         IndexModel(
           Indexes.compoundIndex(
             Indexes.ascending("enrollmentEORINumber"),
             Indexes.ascending("movementType"),
-            Indexes.descending("updated")
```
tudorsonycx, line 172: doesn't matter if the index is ascending or descending; mongo will simply walk through it in reverse if needed, so only having an ascending index on `updated` is enough

